PR #3, opened 2018-09-11
Add "RollingOut" condition to Application

### LINKED ISSUE #8: Add a RollingOut condition in Application
The `RollingOut` condition should indicate whether or not an Application is subject to a transition between Releases.

In the case `RollingOut` status is `True`, the message should contain a description of the transition (for example, "Transitioning from 'foo-1' to 'foo-2'").

In the case `RollingOut` status is `False`, the message should inform the current deployed release (for example, "Last deployed 'foo-2'").

Can't foresee a reason where `RollingOut` should be set to `Unknown`.

isutton:
This patch adds the "RollingOut" condition to Application.

This condition should inform an Application object viewer whether a Release is being rolled out, aiming to make it easier for human viewers to understand and diagnose what is happening with their Release process.

The cases the condition reports are:

- The Contender Release has been successfully rolled out, meaning that no Release is being rolled out: Status is `False`, and Message states the active release in English (`Release %q is active`);
- The Contender Release is being rolled out: Status is `True`, and Message states the names of the Releases involved in this transition, in English (`Transitioning from %q to %q`);
- The Contender Release is being rolled out and there's no Incumbent Release: Status is `True`, and Message states this fact with the name of the Release being rolled out (`Rolling out initial release %q`).

One small unrelated change is the removal of `:` from our internal chart cache path, which is not valid on win32.

Closes #8


kanatohodets:
Like the other branch, I rebased this on master to pick up the Travis checks.

pkg/controller/application/application_controller.go
```diff
@@ -280,174 +243,189 @@ func (c *Controller) processApplication(app *shipperv1.Application) error {
...
-		}
-		app.Annotations[shipperv1.AppHighestObservedGenerationAnnotation] = "0"
-		app.Status.Conditions = conditions.SetApplicationCondition(
-			app.Status.Conditions,
-			shipperv1.ApplicationConditionTypeReleaseSynced,
-			corev1.ConditionTrue,
-			"", "",
-		)
-		return nil
+func (c *Controller) wrapUpApplicationConditions(app *shipperv1.Application, rels []*shipperv1.Release) error {
```
kanatohodets: This function is slightly confusing for me: I see it called as an alternate branch to code paths which also set a bunch of conditions, and this func unconditionally sets a set of conditions without doing any kind of analysis.

I think we might be in an awkward middle ground where we neither have a universal "check all the conditions and set them" func that is the last call before every return, nor do we consistently set the conditions in-line with the work. If I understand correctly, `wrapUpApplicationConditions` means "everything was OK, so mark all the conditions to reflect that and then figure out the RollingOut condition". Is that accurate? If so, I think we can probably workshop a better name; alternatively, we can think about trying to have this thing compute all the various conditions (that might need structured errors?).

pkg/controller/application/application_controller.go
```diff
@@ -475,10 +453,10 @@ func (c *Controller) cleanUpReleasesForApplication(app *shipperv1.Application) {
 	// maintain the invariant that we always delete oldest first (rather than
 	// failing to delete A and successfully deleting B and C in an 'A B C'
 	// history).
-	overhead := len(installedReleases) - int(*app.Spec.RevisionHistoryLimit)
+	overhead := len(installedReleases) - int(*app.Spec.RevisionHistoryLimit) + 1
```
kanatohodets: I think this breaks rollouts when revisionHistoryLimit is 1, since the incumbent gets nuked.

This led to an Application status as following (unrelated conditions removed):
```
status:
  conditions:
  - lastTransitionTime: 2018-09-11T21:32:17Z
    message: Transitioning from "web-serving-excellence-29e5b89b-0" to "web-serving-excellence-4508dfb0-0"
    status: "True"
    type: RollingOut
  history:
  - web-serving-excellence-4508dfb0-0
```

I didn't drop a line off the bottom, the contender was the only member of history (the incumbent had been deleted).

This is my bad for not leaving a note about the slightly suspicious "+1".

test/e2e/e2e_test.go
```diff
@@ -366,8 +366,8 @@ func newFixture(ns string, t *testing.T) *fixture {
 }
 
 func (f *fixture) targetStep(step int, relName string) {
-	patch := fmt.Sprintf(`{"spec": {"targetStep": %d}}`, step)
-	_, err := shipperClient.ShipperV1().Releases(f.namespace).Patch(relName, types.MergePatchType, []byte(patch))
+	patch := fmt.Sprintf(`[{ "op": "replace", "path": "/spec/targetStep", "value": %d }]`, step)
```
kanatohodets: I know we talked about this, but I forgot -- my understanding was that only **Strategic**Merge was unsupported for CRDs, but Merge patches are OK. The reason I think this is because this is how I patch things with kubectl:

```
$ kubectl patch rel web-serving-excellence-f4ef7212-0 --type=merge -p '{"spec": {"targetStep": 1}}' 
release.shipper.booking.com/web-serving-excellence-f4ef7212-0 patched
```

And when I run that with a high log level, this is the API call that `kubectl` is making:
```
I0912 00:10:34.003942   34637 request.go:897] Request Body: {"spec": {"targetStep": 1}}
I0912 00:10:34.004028   34637 round_trippers.go:386] curl -k -v -XPATCH  -H "Accept: application/json" -H "Content-Type: application/merge-patch+json" -H "User-Agent: kubectl/v1.11.1 (darwin/amd64) kubernetes/b1b2997" 'https://192.168.99.100:8443/apis/shipper.booking.com/v1/namespaces/default/releases/web-serving-excellence-f4ef7212-0'
```

pkg/controller/application/application_controller.go
isutton: Removing the incumbent is something that shouldn't happen indeed.

Shouldn't we then limit `RevisionHistoryLimit` to at least `2` then? I may be nitpicking here, but I'd be confused if I read `RevisionHistoryLimit` is `1` and see `2` releases in the history.

pkg/controller/application/application_controller.go
isutton: The method was created to support successful return points on the caller, informing that all the conditions that code path might report are clear; it is only executed in the case no error was found during the caller's execution (with the exception of the case where a release has been added and removed (see below). Other than that, each failure return point sets the proper conditions accordingly.

https://github.com/bookingcom/shipper/blob/bf2f0323862ae18b3da7893aeea5dd6832498f77/pkg/controller/application/application_controller.go#L378-L390

The first action point of this, I think, is to document the method accordingly. The other is finding a better name for it, which it'd be nice to hear your suggestions.

pkg/controller/application/application_controller.go
isutton: Reverted this change but adjusted the original test which was erroneously expecting more releases to be deleted than it should. The semantics of `RevisionHistoryLimit` are unchanged; in `.status.history` will be always present the contender plus `.spec.revisionHistoryLimit` past releases. 

The latest patch also includes a safe-guard to have at least one past release present, in the case the user feels creative and specifies values lesser than `1` for `RevisionHistoryLimit`.

test/e2e/e2e_test.go
isutton: This has been reverted.

isutton:
@kanatohodets: I adjusted the behavior for `revisionHistoryLimit` in cc9c761. Can you check it and proceed to approve this PR?

pkg/clusterclientstore/store.go
```diff
@@ -332,5 +342,15 @@ func buildConfig(host string, secret *corev1.Secret, restTimeout *time.Duration)
 		config.KeyData = key
 	}
 
-	return config
+	if insecureSkipTlsVerify, ok := secret.Data["tls.insecure-skip-tls-verify"]; ok {
```
kanatohodets: I think this might have gotten some of #17 mixed in -- understandable, since you need it for your workflow. Mind resetting this file for this PR?

pkg/controller/application/application_controller_utils.go
```diff
@@ -7,47 +7,38 @@ import (
...
-
-	labels := make(map[string]string)
-	for k, v := range app.GetLabels() {
-		labels[k] = v
-	}
-
-	labels[shipperv1.ReleaseLabel] = releaseName
-	labels[shipperv1.AppLabel] = app.GetName()
-	labels[shipperv1.ReleaseEnvironmentHashLabel] = hashReleaseEnvironment(app.Spec.Template)
+	var (
```
kanatohodets: nit: I think this can be nuked, the only 'err' in the function is := defined later on.

PR #4, opened 2018-09-11
Strategy should work with 1 replica

### LINKED ISSUE #7: Strategy checks don't account for "rounding up" pod counts
This breaks strategy progression when we're looking at small pod counts: it's easy for the incumbent to be stable with too many pods.

isutton:
This patch makes sure that the strategy works with one replica as well as many.

The intended behavior was to not complete the strategy right away but march through the strategy specification even if the whole transition capacity has been fulfilled.

Closes #7.

### DIFF
--- a/pkg/chart/cache/cache.go
+++ b/pkg/chart/cache/cache.go
@@ -27,6 +27,7 @@ func (f *fsCache) Fetch(repo, name, version string) (*bytes.Buffer, error) {
 
 	file := fmt.Sprintf("%s-%s.tgz", name, version)
 	path := filepath.Join(f.dir, repo, name, file)
+	path = strings.Replace(path, ":", "_", -1)
 	_, err := os.Stat(path)
 	if err != nil {
 		if os.IsNotExist(err) {
@@ -97,6 +98,7 @@ func (f *fsCache) Store(data []byte, repo, name, version string) error {
 	}
 
 	filename := fmt.Sprintf("%s-%s.tgz", name, version)
+	filename = strings.Replace(filename, ":", "_", -1)
 	chartPath := filepath.Join(familyPath, filename)
 
 	// Atomic rename swap to avoid cache hits against partially-written charts.
@@ -121,6 +123,7 @@ func clean(names ...string) (string, string, string) {
 		// names a nicer read (not http:__blorg.baz.com).
 		names[i] = strings.TrimPrefix(names[i], "https://")
 		names[i] = strings.TrimPrefix(names[i], "http://")
+		names[i] = strings.Replace(names[i], ":", "_", -1)
 		names[i] = strings.Replace(names[i], string(filepath.Separator), "_", -1)
 	}
 	return names[0], names[1], names[2]
--- a/pkg/controller/errors.go
+++ b/pkg/controller/errors.go
@@ -6,7 +6,11 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 )
 
-type MultipleOwnerReferencesError error
+type MultipleOwnerReferencesError string
+
+func (e MultipleOwnerReferencesError) Error() string {
+	return string(e)
+}
 
 func IsMultipleOwnerReferencesError(err error) bool {
 	_, ok := err.(MultipleOwnerReferencesError)
@@ -14,12 +18,16 @@ func IsMultipleOwnerReferencesError(err error) bool {
 }
 
 func NewMultipleOwnerReferencesError(name string, references int) MultipleOwnerReferencesError {
-	return MultipleOwnerReferencesError(fmt.Errorf(
+	return MultipleOwnerReferencesError(fmt.Sprintf(
 		"expected exactly one owner for object %q, got %d",
 		name, references))
 }
 
-type WrongOwnerReferenceError error
+type WrongOwnerReferenceError string
+
+func (e WrongOwnerReferenceError) Error() string {
+	return string(e)
+}
 
 func IsWrongOwnerReferenceError(err error) bool {
 	_, ok := err.(WrongOwnerReferenceError)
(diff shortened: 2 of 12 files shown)


kanatohodets:
I rebased this branch on master to get the TravisCI pipeline engaged.

pkg/controller/strategy/comparisons.go
```diff
@@ -1,11 +1,23 @@
 package strategy
 
-func contenderCapacityComparison(achieved uint, desired uint) bool {
-	return achieved >= desired
+func contenderCapacityComparison(achievedPercentage, desiredPercentage, totalNumberOfPods uint) bool {
+	// Consider that contender capacity has achieved its desired state if
```
kanatohodets: I think it might make sense to explain here and in the `incumbentCapacityComparison` about our "pod math", where we always round up to the nearest whole pod (and so 10% desired capacity of 1 pod is 1 pod).

pkg/controller/strategy/controller_test.go
```diff
@@ -30,125 +30,142 @@ func init() {
 }
 
 func TestContenderReleasePhaseIsWaitingForCommandForInitialStepState(t *testing.T) {
-	f := newFixture(t)
+	for _, i := range []uint{1, 10} {
```
kanatohodets: How would you feel about having a global slice with the pod counts to test that all the test cases share? That way we can put a comment there explaining why we think these specific values should be tested (mentioning the "pod math" bits).

test/e2e/e2e_test.go
```diff
@@ -268,13 +267,20 @@ func TestNewApplicationVanguard(t *testing.T) {
 	}
 }
 
-func TestRolloutVanguard(t *testing.T) {
+func TestNewApplicationVanguardMultipleReplicas(t *testing.T) {
+	testNewApplicationVanguard(4, t)
```
kanatohodets: Also probably worth putting a note as to why we have these distinct cases (1 pod vs 4 pods).

pkg/controller/strategy/comparisons.go
```diff
@@ -1,11 +1,23 @@
...
+	return achievedPercentage >= desiredPercentage
 }
 
-func incumbentCapacityComparison(achieved uint, desired uint) bool {
-	return achieved <= desired
+func incumbentCapacityComparison(achievedPercentage, desiredPercentage, totalNumberOfPods uint) bool {
+	// Consider that the incumbent capacity has achieved its desired state if
+	// the end game is one pod and we already have it, as long as the desired
+	// is smaller than 100.
+	if totalNumberOfPods == 1 && achievedPercentage == 100 && desiredPercentage < 100 {
```
kanatohodets: I don't totally understand the  `desired < 100` check; my first impression is that it is a bug, since it might say its OK to proceed when we should actually be scaling down to 0. 

Reasoning: if `desiredPercentage` is 0 on a 1 pod deployment, and we (for whatever reason) haven't made any progress turning off that pod, I think this will say that everything is correct, when really it should say that we're not there yet. I think the correct test might be: `totalPods==1 && achievedPercent == 100 && desiredPercentage > 0` for the incumbent as well.

pkg/controller/errors.go
```diff
@@ -6,28 +6,36 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 )
 
-type MultipleOwnerReferencesError error
+type MultipleOwnerReferencesError string
```
kanatohodets: Is this kind of a drive-by fix for this PR, or was it required to make things work? It doesn't seem related to the rest at first glance (our wide-spread bugs with error types notwithstanding).

isutton: Yes, it was a drive-by, but to fix the bug where `shouldRetry` is always evaluated to `false` (see below).

https://github.com/bookingcom/shipper/blob/3015272fe0ec4b28bf67fdad0279d8e6f4e120ec/pkg/controller/installation/installation_controller.go#L187-L190

I'm fine to revert the related changes and create another PR for it.

kanatohodets: Ok cool, no need to split it. I'll grab this in the structured errors work either way.

kanatohodets:
Reflecting our chat today: we understand now that the bug is with any odd numbered replica count on the incumbent. 

Reasoning: if the incumbent has 5 pods, and we are trying to scale it to 50% of its full capacity (2.5 pods), we will scale it to 3 pods (2.5 rounded up to nearest whole pod). As a result, the capacity target object will report that the achieved capacity is 3/5, which is larger than what the strategy thinks the incumbent should be at (2.5), and the rollout gets stuck waiting for the incumbent to scale down.

pkg/chart/cache/cache.go
```diff
@@ -121,6 +123,7 @@ func clean(names ...string) (string, string, string) {
 		// names a nicer read (not http:__blorg.baz.com).
 		names[i] = strings.TrimPrefix(names[i], "https://")
 		names[i] = strings.TrimPrefix(names[i], "http://")
+		names[i] = strings.Replace(names[i], ":", "_", -1)
```
ksurent: I'm assuming these are windows–specific?

pkg/controller/traffic/pod_label_shifter.go
```diff
@@ -217,13 +218,15 @@ func calculateReleasePodTarget(releasePods int, releaseWeight uint32, totalPods
 	if totalWeight == 0 {
 		targetPercent = 0
 	} else {
-		targetPercent = float64(releaseWeight) / float64(totalWeight)
+		targetPercent = float64(releaseWeight) / float64(totalWeight) * 100
 	}
 	// Round up to the nearest pod, clamped to the number of pods this release has.
-	targetPods := int(
+	targetPods := int(replicas.CalculateDesiredReplicaCount(uint(totalPods), float64(targetPercent)))
```
ksurent: `targetPercent` is already defined as `float64`.

pkg/chart/cache/cache.go
ksurent: Nvm, found it.

PR #19, opened 2018-09-19
all: v1 -> v1alpha1

kanatohodets:
When I created the types initially, I was unfamiliar with the
significance of picking a particular API version. As a result, I picked
'v1'. This is not where Shipper is at the moment: we firmly expect some
of our APIs to grow and evolve, potentially drastically, as it matures.

One editorial decision: 'shipperv1' was renamed to 'shipper', not
'shipperv1alpha1'. This is consistent with K8s controllers as far as
I can tell.

Once we merge this, I'm happy to rebase the outstanding PRs on top so this doesn't inflict painful work for everyone.


kanatohodets:
On second thought, we can't yet roll this out -- we need to be running Kubernetes 1.11 in order to do the migration without losing data.

PR #25, opened 2018-10-04
WIP: Introduce RFCs directory

isutton:
This PR contains the README for the rfcs directory. The idea is to, once we're happy with the outline, split it with relevant RFCs documents and discuss about them separately, each into its own PR.

Perhaps in this PR we can discuss also about the format of this RFCs. One idea is copy some prior-art like Python's PEP-1.

### DIFF
--- a/rfcs/README.md
+++ b/rfcs/README.md
@@ -1,6 +1,8 @@
 
 # Roadmap to Open Source Release
 
+This document contains the outline of our Roadmap brainstorm session. The idea is to transform this outline into different RFCs and discuss them as separate Pull Requests.
+
 ## Communication Channels
 
 ### Web Presence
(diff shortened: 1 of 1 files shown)


rfcs/README.md
```diff
@@ -0,0 +1,75 @@
+
+# Roadmap to Open Source Release
```
kanatohodets: nit: probably rephrase because we're already 'released as open source', this is more about when we tag one and try to promote it in the ecosystem.

rfcs/README.md
```diff
@@ -0,0 +1,75 @@
...
+        - **Use cases** helps defining the release workflow.
+        - **Release workflow diagram** helps outlining required tests.
+    - Operations manual
+        - How to install Shipper?
+        - How to manage Shipper?
+            - How do I include a cluster?
+            - How do I cordon a cluster?
+            - How do I drain a cordoned cluster?
+    - User manual
+        - Where do I look for problems
```
icanhazbroccoli: Suggested edits:
- Where do I look for answers
- Where do I look for existing problems

Just for the sake of reducing the ambiguity :-)

PR #26, opened 2018-10-08
E2E test when a strategy is moving backwards refs #13

icanhazbroccoli:
This commit adds 2 new e2e test cases where a rollout strategy is
being tested against a revert scenario. We are testing 2 distinct
scenarios: a brand new application and an existing one rolling
forward. In both cases we are using the existing vanguard strategy:
0 | 50/50 | full-on. New tests are moving the release from step 0
towards 50/50 and revert immediately.


icanhazbroccoli:
❤️❤️❤️

PR #27, opened 2018-10-09
WIP: Off-the-shelf-charts

isutton:


rfcs/999-off-the-shelf-charts.adoc
```diff
@@ -0,0 +1,42 @@
...
+To overcome that limitation, we want to introduce the field `.spec.template.loadBalancer` field in the Application CRD; this field will be used to propagate relevant load balancing configuration to Shipper.
+
+The following listing shows an Application manifest with the `.spec.template.loadBalancer`:
+
+..spec.template.loadBalancer example
+[source,yaml]
+----
+spec:
+  loadBalancer:     # <1>
+    kubernetes:     # <2>
```
kanatohodets: Something I was pondering: we should make this key as specific as we can in order to have some forward compatibility. For example, k8s might introduce some awesome new load balancing thing (IngressNG, for example), and if we name this key “kubernetes”, we won’t be able to add an implementation for that without inventing another vanilla Kubernetes key.

 So a really rough sketch might be:
```
loadBalancer:
  kubernetes:
    core/v1/service:
      name: foobar
      selector: {}
```

Looking at it, you might not need ‘kubernetes’ as a level anymore, since GVK is unambiguous. 

Same principle applies to Istio if they change their APIs around. However, it does make our syntax more verbose, which isn’t super exciting.

kanatohodets:
Closing, this is superseded by the more targeted workaround discussed in https://github.com/bookingcom/shipper/pull/32. We discovered discussing this issue that we probably don't have enough information or experience to design the API for that 'loadbalancer' key just yet.

PR #29, opened 2018-10-11
RFC: Introduce shipperctl command line utility

isutton:


osdrv (approved):

osdrv (dismissed):

rfcs/999-shipperctl-admin.adoc
```diff
@@ -0,0 +1,94 @@
...
+|`-f, --file`
+| Shipper cluster configuration file to be applied
+
+|`--dry-run`
+| Print out what actions it would do, according to the given cluster configuration file
+
+|`-n, --namespace`
+| Shipper system namespace to be created in all clusters
+
+|`-k, --kubeconfig`
```
kanatohodets: I'd avoid using `-k`, because it isn't a shortcut that `kubectl` uses for `kubeconfig`.

rfcs/999-shipperctl-admin.adoc
```diff
@@ -0,0 +1,94 @@
...
+
+=== Set Up the Application Cluster
+
+* Create the Namespace based on the `namespace` argument on the application cluster
+* Create a service account called `shipper` in the namespace on the application cluster
+* Create a ClusterRoleBinding to bind the Shipper service account to the `cluster-admin` ClusterRole on the application cluster
+
+=== Connect the Clusters Together
+
+* Create the cluster Secret on the management cluster
```
kanatohodets: Just to add a missing bit: the contents of the Secret created on the Management cluster for this Application cluster should be the Service Account Token (which will be a secret called 'shipper-token-$hash' in the shipper namespace). This typically includes a few fields: `ca.crt`, `namespace`, and `token`. Effectively we can copy this token secret into the management cluster, and change the type to `Opaque`, and add the 'checksum' field.

kanatohodets (dismissed):
LGTM! Couple of minor comments to flesh things out

osdrv (dismissed):

rfcs/999-shipperctl-admin.adoc
```diff
@@ -0,0 +1,94 @@
...
+* Register CRDs on the management cluster
+* Create a service account called `shipper` in the Shipper namespace on the management cluster
+* Create a ClusterRole called `shipper-admin` to manage Shipper resources on the management cluster
+* Create a ClusterRoleBinding to bind the shipper service account to the shipper-admin role on the management cluster
+* Create a Role called `secret-reader` in the Shipper namespace that gives permission to read secrets on the management cluster
+* Create a RoleBinding to bind the Shipper service account to the `secret-reader` role
+
+=== Set Up the Application Cluster
+
+* Create the Namespace based on the `namespace` argument on the application cluster
```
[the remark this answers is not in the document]
parhamdoustdar: Great point. I was going to add it to the RFC, but I saw that it is already mentioned in the second use case:

> The second use case is to update the Cluster objects, or to re-do any of the steps that are necessary to add this cluster. For example, if a service account is already set up for Shipper, this command will skip that step. This makes this command idempotent, and allows it to be used for git-ops or any sort of pipeline.

Is this what you meant?

osdrv (dismissed):

PR #30, opened 2018-10-16
Implemented e2e tests for aborts refs #11

icanhazbroccoli:
This commit extends the end-to-end test suite with 2 new test cases:
one ofr a new application and the other one for an existing one.
In both cases, we remove the current active release object and watch
what's happening. In the case of a new application we expect the system
to revert everything to step 0 and wait for a command, whereas an
existing application, which is being rolled forward, should pick up the
previous existing release and move full-on to the last step of this
release.

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -491,7 +491,7 @@ func TestRolloutMovingStrategyBackwards(t *testing.T) {
 	}
 }
 
-// TestNewApplicationAbort emulates a brand new application rolout.
+// TestNewApplicationAbort emulates a brand new application rollout.
 // The rollout strategy includes a few steps, we are creating a new release,
 // Next, we are moving 1 step forward (50% of the capacity and 50% of the
 // traffic) and delete the release. The expected behavior is:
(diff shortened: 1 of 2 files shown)


icanhazbroccoli:
Automatic check failed, and same failures are reproducible on my local setup. I suspect some bit of non-determinism to be present here. Investigating, will update this thread asap.

icanhazbroccoli:
The failure seems to be related to a race condition happening on resync: increasing the sync period to 2minutes reduces the amount of test failures down to 2–5%.

icanhazbroccoli:
@isutton found a race condition in a case when one removes a release and then immediately creates a new one. Release owns a set of children objects: capacity target, traffic target and installation target. When scheduler creates the corresponding objects for a new release, it always checks if they already exist. If they do, scheduler says "my job here is done even though i did nothing". So far so good, but the objects are eventually being garbage-collected by Kubernetes as with a removed release they are orphans now. The new objects are not being created at all and this is a forever-stall situation. We changed the behaviour of scheduler to returning an error whenever an object exists and relying on the upcoming retry attempt to go clean.

test/e2e/e2e_test.go
```diff
@@ -491,6 +491,169 @@ func TestRolloutMovingStrategyBackwards(t *testing.T) {
 	}
 }
 
+// TestNewApplicationAbort emulates a brand new application rolout.
```
isutton:
```suggestion
// TestNewApplicationAbort emulates a brand new application rollout.
```

PR #33, opened 2018-10-26
ci: add -e to test.sh

kanatohodets:
'go test' exiting non-zero was not causing the build to fail


kanatohodets:
The checks are failing because the underlying tests really are failing, but this was previously masked by a missing `set -e` in the test.sh script.

kanatohodets:
I used admin override to merge this despite the failures

PR #35, opened 2018-10-29
RFC: Assess chart health before creating a release in Application Controller

isutton:
This PR introduces the RFC to access chart health before creating a release in Application Controller.

The idea is to slim down the Schedule Controller, moving any chart processes from Schedule Controller to the Application Controller, propagating any sort of chart issue to the Application's conditions.


kanatohodets:
We had a discussion and are going to rethink this effort, likely in the direction of merging the Schedule and Strategy controllers into a Release controller.

PR #37, opened 2018-10-30
RFC: Introduce Release Controller

isutton:


rfcs/999-introduce-release-controller.adoc
```diff
@@ -0,0 +1,126 @@
...
+encoded directly in the Release conditions right after acting.
+
+==== Composable act operations
+
+The Web Services Middlewares approach can function as a simplified finite state
+machine, where at any point in the execution the processing can be interrupted
+by skipping the next state returning immediately to the caller (this is also
+called link:https://fsharpforfunandprofit.com/rop[Railway Oriented Programming]
+(here is a link:https://github.com/dc0d/rop[Go implementation of ROP] for
+reference).
```
kanatohodets: Probably we want to be a little careful about adopting super high power abstractions. I think we can probably navigate this just by paying attention to the high level APIs of the "schedule" and "run strategy" modules.

rfcs/999-introduce-release-controller.adoc
```diff
@@ -0,0 +1,126 @@
+= 999 Introduce Release Controller
+Igor Sutton <igor.sutton@booking.com>
+1970-01-01
```
kanatohodets: I think we should either kill this metadata or have it be "real". Bit funky to have a bunch of '999-foo' and "standards track" things.

PR #32, opened 2018-10-25
Implement a Workaround For Labels Injected By Helm Into Charts, and Make shipper-lb Optional

parhamdoustdar:

PR #39, opened 2018-11-01
Workaround for labels: "shipper-lb" is optional, "release" is a no-go #32

icanhazbroccoli:
This commit introduces a breaking change in the installation controller
logic. Shipper expects "shipper-lb": "production" label to present in a
service object in order to realise the load-balancing element. With this
commit a user might drop this tag if there is only 1 service in the
chart.
Release object is something that helm puts in by default and is
something that breaks shipper capacity/traffic shifting. If a user
provides a magical application-level label:
"shipper-enable-helm-release-workaround":"1", it would inform the
installer to remove the release label and proceed with the normal
shifting strategy. Otherwise it's an error and it will bail out.


icanhazbroccoli:
Based on @kanatohodets feedback, I'm planning to give installer some more love and beautify it a bit.

PR #43, opened 2018-11-08
Implemented a workaround for off-the-shelf charts #32

icanhazbroccoli:
This commit introduces a few changes to the installer behavior:
  * If there is only 1 service object in a chart but it contains no
  shipper-lb label, installer will pass it as is: there is not that much
  to disambiguate. If there is more than 1 service, it will raise an error
  * If a service selector contains selector label `release`, shipper
  will raise an error because it breaks the traffic shifting logic. The
  user can tell the system to remove this label if present by providing
  `enable-helm-release-workaround` application metadata label. The
  actual value is never checked, can be either "true" or "enabled".


pkg/controller/installation/installer.go
```diff
@@ -162,23 +162,60 @@ func (i *Installer) patchService(
...
+			// This selector label is native to helm-bootstrapped charts.
+			// In order to make it work the shipper way, we remove the
+			// label and proceed normally. With one little twist: the user
+			// has to ask shipper to do it explicitly.
+			if relName == i.Release.Name {
+				delete(s.Spec.Selector, shipperv1.HelmReleaseLabel)
+			}
+		} else {
+			return nil, controller.NewInvalidChartError(
+				fmt.Sprintf("The chart contains %q label. This will break shipper"+
```
parhamdoustdar: I like this error, but maybe we can make it clearer? Something like:

```
fmt.Sprintf("The chart contains %q label. This will break shipper"+
 					" traffic shifting logic. Consider adding the workaround label %q: true to your Application object",
 					shipperv1.HelmReleaseLabel, shipperv1.HelmWorkaroundLabel))
```

icanhazbroccoli: Fixed in the recent commit. Thank you!

pkg/controller/installation/installer.go
```diff
@@ -253,13 +290,23 @@ func (i *Installer) installManifests(
 
 	// Keep decoded and unstructured objects around until we are ready to
 	// process them.
-	var renderedObjects []struct {
-		unstructured *unstructured.Unstructured
-		decoded      runtime.Object
-	}
+	// var renderedObjects []struct {
```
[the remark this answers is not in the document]
icanhazbroccoli: Fixed in the recent commit. Thank you!

pkg/controller/installation/installer_test.go
```diff
@@ -303,3 +304,191 @@ func TestInstallerBrokenChartContents(t *testing.T) {
 		t.Fatal("installRelease should fail, invalid k8s objects")
 	}
 }
+
+func TestInstallerSingleServiceNoLB(t *testing.T) {
```
kanatohodets: I think it might make sense to check the contents of the created Services, so we can check that they have the Selectors that we expect (unmodified if a secondary Service and the primary has the 'shipper-lb' label; modified if its the only Service or it has the shipper-lb prod label, etc.)

pkg/controller/installation/installer.go
```diff
@@ -162,23 +162,61 @@ func (i *Installer) patchService(
...
-		// If we reach this point, it is fine to panic() since there's nothing
-		// much to do if we don't have the application label.
-		panic(fmt.Sprintf("Programmer error, label %q should always be present", shipperv1.AppLabel))
+
+	requiredLabels := []string{shipperv1.AppLabel}
+	for _, label := range requiredLabels {
+		if _, ok := labelsToInject[label]; !ok {
+			// If we reach this point, it is fine to panic() since there's nothing
+			// much to do if we don't have the mandatory labels.
+			panic(fmt.Sprintf("Programmer error, label %q should always be present", label))
```
kanatohodets: I think what you wrote elsewhere (returning an error in the case of a missing label) makes more sense than `panic()`-ing, even if this is what the code was doing before.

pkg/controller/installation/installer_test.go
icanhazbroccoli: We do check the contents of Service objects in validateServiceCreateObject and in validateDeploymentCreateObject + this commit enforces a precise label check: https://github.com/bookingcom/shipper/pull/43/files#diff-2e47c909741e399c1373620e5e4f379aL170

isutton:
>     * The actual value is never checked, can be either "true" or "enabled".

I think the value should be always checked, if present. It'd be extremely annoying if we say "false" there and the feature is still turned on.

Note: I'm writing this without having read the whole thread, will update this in a few if my concern was raised somewhere.

pkg/controller/installation/installer.go
```diff
@@ -154,38 +154,77 @@ func (i *Installer) patchDeployment(
...
-	if !ok {
-		// If we reach this point, it is fine to panic() since there's nothing
-		// much to do if we don't have the application label.
-		panic(fmt.Sprintf("Programmer error, label %q should always be present", shipper.AppLabel))
+) (runtime.Object, error) {
+
+	requiredLabels := []string{shipper.AppLabel}
+	for _, label := range requiredLabels {
+		if _, ok := labelsToInject[label]; !ok {
+			// If we reach this point, it is fine to panic() since there's nothing
```
isutton: I think this comment is now outdated.

pkg/controller/installation/installer.go
```diff
@@ -154,38 +154,77 @@ func (i *Installer) patchDeployment(
...
 
 	// We are interested only in patching Services properly identified by our specific label
-	if lbValue, ok := s.Labels[shipper.LBLabel]; ok && lbValue == shipper.LBForProduction {
+	if lbValue, ok := labels[shipper.LBLabel]; ok && lbValue == shipper.LBForProduction {
 		s.Spec.Selector[shipper.PodTrafficStatusLabel] = shipper.Enabled
 	}
-	return s
+
+	if relName, ok := s.Spec.Selector[shipper.HelmReleaseLabel]; ok {
+		if _, ok := i.InstallationTarget.Labels[shipper.HelmWorkaroundLabel]; ok {
```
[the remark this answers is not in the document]
kanatohodets: This looks fixed in the latest revision: 'true' enables it, 'false' disables it, any other value triggers an error saying "you should put either true or false there", which sounds good to me.

pkg/controller/installation/installer.go
kanatohodets: Fixed in https://github.com/bookingcom/shipper/pull/43/commits/bbb4bac9ac90257461bb74a137f4fbb29c2d5a8b

PR #44, opened 2018-11-14
Propagate annotations from application object to release object

irajdeep:
Currently, shipper only propagates labels from application to release object but not annotations.
Label values have a limitation on the characters that can be used, 
For ex: if one cannot use an e-mail address as a label value because `@` character is not allowed, in such cases annotations should be used


kanatohodets:
Sorry for the hold up. Here are my thoughts on this:

If we want to add this, we should go all the way in emulating the Deployment control's relationship with ReplicaSets, and thus include the logic which pushes the annotations back to the Application when an abort happens (latest Release deleted).

I'm not totally sure about the use case yet, though. I don't have a good feeling for whether the Deployment<->ReplicaSet annotation logic is a useful thing that we should emulate. We haven't really done much talking about what kinds of things folks might want to do with annotations on Applications.

kanatohodets:
Closing for now, feel free to re-open if you want to start the conversation again.

PR #45, opened 2018-11-19
Introduced shipperctl admin clusters apply for setting up and updating cluster configuration

parhamdoustdar:
This pull request introduces the `shipperctl` command line utility. For now, it only has the `shipperctl clusters apply` command, which accepts a cluster configuration and uses it to set up management and application clusters.

Other things I did that had nothing to do with the original intent of this branch were to upgrade the version of `dep` we're using to manage our dependencies, and to exclude all the errors beginning with `vendor` from the `gometalinter` output. These were done to make our CI pipeline pass.


.travis.yml
```diff
@@ -15,7 +15,7 @@ env:
 before_install:
 - sudo snap install microk8s --classic --channel=1.11/stable
 - curl -Lo $GOPATH/bin/kubectl https://storage.googleapis.com/kubernetes-release/release/v1.11.4/bin/linux/amd64/kubectl
-- curl -Lo $GOPATH/bin/dep https://github.com/golang/dep/releases/download/v0.4.1/dep-linux-amd64
+- curl -Lo $GOPATH/bin/dep https://github.com/golang/dep/releases/download/v0.5.0/dep-linux-amd64
```
icanhazbroccoli: It might make sense to check if $GOBIN is set and points to a different location

cmd/shipperctl/cmd/apply.go
```diff
@@ -0,0 +1,371 @@
...
+	"gopkg.in/yaml.v2"
+
+	"github.com/bookingcom/shipper/cmd/shipperctl/config"
+	"github.com/bookingcom/shipper/cmd/shipperctl/configurator"
+	"github.com/bookingcom/shipper/pkg/crds"
+)
+
+var applyCmd = &cobra.Command{
+	Use:   "apply",
+	Short: "Set up management and application clusters to work with Shipper",
```
icanhazbroccoli: I appreciate these descriptions, it gives an idea what the package does. I'm also wondering if it makes any sense to write regular package godoc-friendly comments so the automatically built documentation will include it right away?

cmd/shipperctl/configurator/cluster.go
```diff
@@ -0,0 +1,312 @@
...
+	newSecret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      cluster.Name,
+			Namespace: newNamespace,
+			Annotations: map[string]string{
+				shipperv1.SecretChecksumAnnotation: hex.EncodeToString(hash.Sum(nil)),
+			},
+			OwnerReferences: []metav1.OwnerReference{
+				metav1.OwnerReference{
+					APIVersion: "shipper.booking.com/v1",
```
icanhazbroccoli: Does it make any sense to keep it in a constant?

.travis.yml
kanatohodets: Good call, but let's do it in another PR; the dep version change is already basically a drive-by just to work around a dep bug.

cmd/shipperctl/configurator/cluster.go
kanatohodets: Particularly given that it's the wrong version! This got me thinking where one might find a constant with this available, and as it turns out, I don't think there is such a thing anywhere. I guess we can drop one into apis/v1alpha somewhere.

I think it'd be ok to just change the string to be 'v1alpha1' in this PR, and let's try to fix all the version strings in another round; there are many more than this one IIRC.

cmd/shipperctl/configurator/cluster.go
```diff
@@ -0,0 +1,312 @@
...
+	"encoding/hex"
+	"hash/crc32"
+
+	apiextensionv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+
+	apiextensionclientset "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+
+	shipperclientset "github.com/bookingcom/shipper/pkg/client/clientset/versioned"
+
+	shipperv1 "github.com/bookingcom/shipper/pkg/apis/shipper/v1alpha1"
```
kanatohodets: `shipper` package alias instead of `shipperv1` here and throughout the file

cmd/shipperctl/cmd/apply.go
```diff
@@ -0,0 +1,371 @@
...
+	shipperManagementClusterRoleBindingName  = "shipper"
+	shipperApplicationClusterRoleBindingName = "shipper-cluster-admin"
+	shipperSecretReaderRoleName              = "secret-reader"
+	shipperSecretReaderRoleBindingName       = "shipper"
+)
+
+func init() {
+	applyCmd.Flags().StringVarP(&configFile, "file", "f", "clusters.yaml", "config file")
+	applyCmd.Flags().StringVar(&kubeConfigFile, "kube-config", "~/.kube/config", "the path to the Kubernetes configuration file")
+	applyCmd.Flags().StringVarP(&shipperSystemNamespace, "shipper-system-namespace", "n", "shipper-system", "the namespace where Shipper is running")
```
kanatohodets: There's a constant for this in `pkg/apis/shipper/v1alpha`: `ShipperNamespace`.

cmd/shipperctl/configurator/cluster.go
parhamdoustdar: Okay, will keep this as is, so that it gets fixed after merging pull request #49.

icanhazbroccoli:
Just a suggestion. I ran the utility with no command argument:
```
./shipperctl -f ~/workspace/kubernetes/clusters.yaml
flag provided but not defined: -f
Usage of ./shipperctl:
  -alsologtostderr
        log to standard error as well as files
  -log_backtrace_at value
        when logging hits line file:N, emit a stack trace
  -log_dir string
        If non-empty, write log files in this directory
  -logtostderr
        log to standard error instead of files
  -stderrthreshold value
        logs at or above this threshold go to stderr
  -v value
        log level for V logs
  -vmodule value
        comma-separated list of pattern=N settings for file-filtered logging
```

I assume, it would be helpful to output available commands and mention it is expecting it at all.

PR #54, opened 2018-11-27
Add missing OpenAPI chart validation

isutton:


crd/Release-crd.yaml
```diff
@@ -32,97 +32,102 @@ spec:
...
+                  properties:
+                    name:
+                      type: string
+                    version:
+                      type: string
+                    repoUrl:
+                      type: string
+                sidecars:
                   type: array
+                  nullable: true
```
ksurent: This should probably be accompanied by `omitempty` for the corresponding type?

kanatohodets:
Rebased on master and added the validation changes to the CRD structs that `shipperctl` uses. I think this is ready to merge; Oleg and I will review together Monday.

Issue #46, opened 2018-11-21
Create a Shipperctl Command to Validate Helm Charts


parhamdoustdar:
Right now, there is no quick and easy way to figure out what Helm charts work with Shipper. It would be great if we could provide a command to check a chart against our constraints.

Here are the benefits for this approach:

- Users can validate the charts they've written before installing it
- Users can validate Helm charts in the stable repo before using them

One thing that needs mentioning here is that the scope of this command should only be to check the chart against Shipper constraints. We're not telling users if their objects will install or work as expected – we're only telling them that it would pass Shipper's rudimentary checks.

icanhazbroccoli:
A PR is ready to be reviewed: https://github.com/bookingcom/shipper/pull/59

PR #59, opened 2018-12-12
shipperctl validate chart command #46

icanhazbroccoli:
This pull request is aiming to implement a new shipperctl command which will perform a quick probe of a helm chart and report potential issues.


icanhazbroccoli:
Hey folks, this PR is now in an i-think-it's-ready-for-a-review state. Please give it a quick look and let me know your impressions @kanatohodets @parhamdoustdar @isutton

pkg/chart/validate.go
```diff
@@ -0,0 +1,92 @@
...
+	lbs := make([]*corev1.Service, 0)
+
+	for _, decodedObj := range decoded {
+		if svc, ok := decodedObj.(*corev1.Service); ok {
+			services = append(services, svc)
+			if lbValue, ok := svc.Labels[shipper.LBLabel]; ok && lbValue == shipper.LBForProduction {
+				lbs = append(lbs, svc)
+				if len(lbs) > 1 {
+					return fmt.Errorf(
+						fmt.Sprintf("Object %#v contains %q label, but %#v claims"+
```
ksurent: `Errorf` does formatting, don't need `fmt.Sprintf`.

pkg/chart/validate.go
```diff
@@ -0,0 +1,92 @@
...
+							" it is the production LB. This looks like a misconfig:"+
+							" only 1 service is allowed to be the production LB.",
+							decodedObj, shipper.LBLabel, lbs[0]))
```
ksurent: Nit: you probably don't want to dump the entire object here, GVK+name should be enough.

pkg/controller/installation/errors.go
```diff
@@ -69,6 +69,10 @@ type RenderManifestError error
 
 type IncompleteReleaseError error
 
+type FetchChartError error
```
ksurent: This is likely not what you want, it's **not** type aliasing. We have a bunch of places like this in the code and they all should be fixed.

We have a custom build of staticcheck that catches this, please use it and consider adding it to CI.

icanhazbroccoli:
@ksurent I really appreciate your feedback. Very much helpful!

kanatohodets:
Build should start working once rebased on master (codegen needed updating)

cmd/shipperctl/cmd/validate/helm_chart.go
```diff
@@ -0,0 +1,38 @@
...
+
+func runValidateHelmChartCommand(cmd *cobra.Command, args []string) error {
+	for _, chartPath := range args {
+		// TODO: make it understand remote chart URLs
+		// use chart.downloadChart/3
+		chart, loadErr := chartutil.Load(chartPath)
+		if loadErr != nil {
+			return fmt.Errorf("Failed to load chart under path %q: %s", chartPath, loadErr.Error())
+		}
+		if validateErr := shipperchart.Validate(chart); validateErr != nil {
```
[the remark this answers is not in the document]
luisdavim: I think that both validations are useful, I would want to validate that the `app.yaml` and the charts referenced in it are ok but I would also a quick way to check if an existing chart would work with shipper, in the second case using the default values should be ok.

icanhazbroccoli:
Closing this PR due to an uncertainty regarding it's applicability and usability.

PR #60, opened 2018-12-13
Validating webhook

### LINKED ISSUE #51: YAML problems on the Application object are not surfaced
Today a user inserted the following Application object into a management cluster:

```yaml
apiVersion: shipper.booking.com/v1alpha1                                                                                                                                                                                                                                                                                                                                                               
kind: Application                                                                                                                                                                                                                                                                                                                                                                                
metadata:                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                    
  name: some-application                                                                                                                                                                                                                                                                                                                                                                                 
  namespace: some-namespace                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                
spec:                                                                                                                                                                                                                                                                                                                                                                                            
  revisionHistoryLimit: 10                                                                                                                                                                                                                                                                                                                                                                       
  template:                                                                                                                                                                                                                                                                                                                                                                                      
    chart:                                                                                                                                                                                                                                                                                                                                                                                       
      name: cool-chart                                                                                                                                                                                                                                                                                                                                                                                 
      repoUrl: https://internal-chart-repo.example.com/charts                                                                                                                                                                                                                                                                                                                                            
      version: 12                                                                                                                                                                                                                                                                                                                                                                            
    clusterRequirements:                                                                                                                                                                                                                                                                                                                                                                         
      capabilities:                                                                                                                                                                                                                                                                                                                                                                              
      - cloudlb                                                                                                                                                                                                                                                                                                                                                                                      
      regions:                                                                                                                                                                                                                                                                                                                                                                                   
      - name: us-east1                                                                                                                                                                                                                                                                                                                                                                             
        replicas: 1                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       
    sidecars:
      foo:
        enabled: false                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            
    strategy:                                                                                                                                                                                                                                                                                                                                                                                    
      steps:                                                                                                                                                                                                                                                                                                                                                                                     
      - capacity:                                                                                                                                                                                                                                                                                                                                                                                
          contender: 1                                                                                                                                                                                                                                                                                                                                                                           
          incumbent: 100                                                                                                                                                                                                                                                                                                                                                                         
        name: staging                                                                                                                                                                                                                                                                                                                                                                            
        traffic:                                                                                                                                                                                                                                                                                                                                                                                 
          contender: 0                                                                                                                                                                                                                                                                                                                                                                           
          incumbent: 100                                                                                                                                                                                                                                                                                                                                                                         
      - capacity:                                                                                                                                                                                                                                                                                                                                                                                
          contender: 10                                                                                                                                                                                                                                                                                                                                                                          
          incumbent: 90                                                                                                                                                                                                                                                                                                                                                                          
        name: canary                                                                                                                                                                                                                                                                                                                                                                           
        traffic:                                                                                                                                                                                                                                                                                                                                                                                 
          contender: 10                                                                                                                                                                                                                                                                                                                                                                          
          incumbent: 90                                                                                                                                                                                                                                                                                                                                                                          
      - capacity:                                                                                                                                                                                                                                                                                                                                                                                
          contender: 100                                                                                                                                                                                                                                                                                                                                                                         
          incumbent: 0                                                                                                                                                                                                                                                                                                                                                                           
        name: full on                                                                                                                                                                                                                                                                                                                                                                            
        traffic:                                                                                                                                                                                                                                                                                                                                                                                 
          contender: 100                                                                                                                                                                                                                                                                                                                                                                         
          incumbent: 0                                                                                                                                                                                                                                                                                                                                                                           
    values:                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                    
      image:                                                                                                                                                                                                                                                                                                                                                                                     
        repository: internal-docker-registry.example.com                                                                                                                                                                                                                                                                                                                       
        tag: deadbeef                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                     
      ports:                                                                                                                                                                                                                                                                                                                                                                                     
      - 443                                                                                                                                                                                                                                                                                                                                                                                      
      replicaCount: 1        
```

This yielded errors in the Shipper logs like so:

```
E1122 14:54:29.831311       1 reflector.go:205] github.com/bookingcom/shipper/pkg/client/informers/externalversions/factory.go:74: Failed to list *v1alpha1.Application: v1alpha1.ApplicationList: Items: []v1alpha1.Application: v1alpha1.Application: Spec: v1alpha1.ApplicationSpec: Template: v1alpha1.ReleaseEnvironment: Sidecars: []v1alpha1.Sidecar: ReadArrayCB: expect [ or n, but found {, error found in #10 byte of ...|idecars":{"foo":|..., bigger context ... ,"sidecars":{"foo":{"enabled":false}},"strategy":{"steps":[|...
```
(there's an error in the Application yaml: 'sidecars' is a vestigial key that we should delete, and it had contents that our struct doesn't expect)

The Application object had no 'status' written, no events, and no Release created. As a result, the user was really confused about why their rollout wasn't going anywhere.

A few tasks here:

1) remove 'sidecars' from our Application/Release API, Shipper isn't doing sidecar injection and probably shouldn't start without careful consideration.
2) Make sure that YAML problems in the `template` are surfaced in 'status' and events of the Application, ideally also on a Release object.
3) Look into stricter validation using an admission controller or similar.

isutton:
To prevent Kubernetes to accept Shipper objects that do not conform with the schema, this PR introduces a validation webhook exposed so Shipper can review each create and update before it hits etcd.

The "webhook-controller" is started like any other of Shipper's controllers, and also responds to the stop signal that other controllers respect. It is a web-server that has an endpoint configured to ingest AdmissionReview payloads provided by Kubernetes Admission Controller, and return an AdmissionResponse with the validation's result.

## Assets

A couple of assets were introduced to help setting it up (shipperctl integration is currently out of the scope of this PR):

* Makefile
  
  A Makefile was introduced to help build Shipper's related images (shipper and shipper-state-metrics), also to build webhook-controller's key and signed certificate.

* hack/webhook/webhook-create-signed-cert.sh 

  A helper script that creates a private key for the server, and automation around Kubernetes CertificateSigningRequest system, which allows Kubernetes to sign CSRs automatically. At the end of the process, this script creates a Secret in the `shipper-system` namespace containing the webhook-controller's private key and signed certificate to be used by the Shipper when starting the HTTPS server.
  
  For the automatic signing to work in microk8s, the `${SNAP_DATA}/args/kube-controller-manager` must be edited, and the following lines should be added:
  
      --cluster-signing-cert-file=${SNAP_DATA}/certs/ca.crt
      --cluster-signing-key-file=${SNAP_DATA}/certs/ca.key

  This configuration enables a Kubernetes subsystem that signs the CSR once an operator has approved it through `kubectl certificate approve` command.
  
  The original script was found at https://medium.com/ibm-cloud/diving-into-kubernetes-mutatingadmissionwebhook-6ef3c5695f74.

* hack/webhook/webhook-patch-ca-bundle.sh 
  
  A helper script that gathers Kubernetes' CA bundle from Kubernetes itself so it can be expanded in kubernetes/validating-webhook-configuration.yaml
  
* kubernetes/validating-webhook-configuration.yaml
  
  Instructions for the Admission Controller to validate Shipper objects both when creating new objects or updating existing ones. By default, its `failurePolicy` is set to "Fail", meaning that any changes in Shipper objects when Shipper is not being executed will be fail (this seemed like a nice default for me since changes on API master won't modify the current state until Shipper is back online).

* kubernetes/shipper.service.yaml 
  
  Since Shipper's webhook-controller exposes a HTTPS server on port 9443, a Service is required for Kubernetes to locate Shipper's validating webhook. The object created using this manifest is referred in the Validating Webhook Configuration mentioned earlier on.
  
* kubernetes/shipper.deployment.yaml

  This is the equivalent of existing and documented Shipper's deployment, with two particular tweaks: one is informing Kubernetes it exports port 9443, and the other is that a secret containing the signed certificate and server keys, produced by hack/webhook/webhook-create-signed-cert.sh are mounted and available to be used by the webhook-controller.

* Dockerfile.shipper

  Since microk8s ships with Docker 17.03, and it is currently not advised to use microk8s Docker and another more recent Docker on the same box (at the time of this writing at least), the build argument used to specify the FROM image was removed, thus hard-coding the base image in the Dockerfile itself. Additionally, the output directory when building shipper now is ${PROJECT_ROOT}/build.
  
* Dockerfile.shipper-state-metrics

  This asset was renamed from Dockerfile.metrics so it is clear what it does, and also will be useful when streamlining the created Makefile later on or using yet build system. Additionally, the output directory when building shipper now is ${PROJECT_ROOT}/build.

* ci/package.sh

  Changed shipper-state-metrics input Dockerfile to reflect the renaming mentioned earlier.

## Code

* pkg/webhook/controller/webhook_controller.go

  That is a very simple controller that uses the same interface as others: provide a NewController function, that returns a Controller object that responds to the `Run()` method.
  
  The `initializeHandlers()` method installs the validation handler function in '/validate'. 
  
  The `validateHandlerFunc()` method is responsible for the simple validation, which is performed by un-marshalling the AdmissionRequest raw object into the equivalent informed by the Kind informed, and in case the decoding has been successfully performed, an AdmissionResponse with its Allowed field set to true is returned.
   
  The `validateHandlerFunc()` is wrapped by the `adaptHandler()` function to be used in HTTP context (`adaptHandler()` is responsible for extracting the AdmissionReview object from the request and transform the resulting AdmissionResponse into an HTTP response object).

* cmd/shipper/main.go

  A couple of options have been added into shipper:

      -webhook-cert - path to the TLS certificate for the webhook-controller 
      -webhook-key  - path to the TLS private key for the webhook-controller
      -webhook-addr - address to bind the webhook-controller
      -webhook-port - port to bind the webhook-controller

  Additionally, respective fields have been added to the configuration type used by all controllers; added also infra-structure similar to other existing controllers to start the webhook-controller. 

* pkg/apis/shipper/v1alpha1/types.go

  To make Shipper work with the validating webhook just created, it is required to use the `omitempty` Json tag in ReleaseEnvironment.Sidecars, otherwise the Application Controller itself couldn't create a Release.
    
## Testing the Setup

In order to configure an existing microk8s cluster with this setup, one can use the Makefile like the following: 

    $ make certs shipper install

Fixes #51.


Makefile
```diff
@@ -0,0 +1,28 @@
...
+	docker build -f Dockerfile.shipper -t localhost:32000/shipper:latest --build-arg HTTP_PROXY=$(HTTP_PROXY) --build-arg HTTPS_PROXY=$(HTTPS_PROXY) .
+	docker push localhost:32000/shipper:latest
+
+shipper-state-metrics:
+	GOARCH=amd64 CGO_ENABLED=0 GOOS=linux go build -o build/shipper-state-metrics ./cmd/shipper-state-metrics/*.go
+	docker build -f Dockerfile.shipper-state-metrics -t localhost:32000/shipper-state-metrics:latest --build-arg HTTP_PROXY=$(HTTP_PROXY) --build-arg HTTPS_PROXY=$(HTTPS_PROXY) .
+	docker push localhost:32000/shipper-state-metrics:latest
+
+restart:
+	$(KUBECTL) get po -o jsonpath='{.items[*].metadata.name}' | xargs $(KUBECTL) delete po
```
ksurent: `kubectl delete pod --all`?

Dockerfile.shipper-state-metrics
```diff
@@ -1,6 +1,5 @@
-ARG BASE_IMAGE=alpine:3.8
```
ksurent: This is used by internal CI.

Dockerfile.shipper-state-metrics
```diff
@@ -1,6 +1,5 @@
...
-FROM ${BASE_IMAGE}
+FROM alpine:3.8
 LABEL authors="Parham Doustdar <parham.doustdar@booking.com>, Alexey Surikov <alexey.surikov@booking.com>, Igor Sutton <igor.sutton@booking.com>, Ben Tyler <benjamin.tyler@booking.com>"
 RUN apk add ca-certificates
-ADD shipper-state-metrics /bin/shipper-state-metrics
-ENTRYPOINT ["shipper-state-metrics", "-v", "2"]
\ No newline at end of file
+ADD build/shipper-state-metrics /bin/shipper-state-metrics
```
ksurent: This path is used by internal CI.

ci/package.sh
```diff
@@ -22,7 +22,7 @@ echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
 docker build -t bookingcom/shipper:$TRAVIS_COMMIT -f Dockerfile.shipper .
 docker push bookingcom/shipper:$TRAVIS_COMMIT
 
-docker build -t bookingcom/shipper-state-metrics:$TRAVIS_COMMIT -f Dockerfile.metrics .
```
ksurent: This file is used by internal CI.

Dockerfile.shipper-state-metrics
isutton: Given that we want to use the same Dockerfile in both environments, I think we should be more conservative with Docker features (for example, using microk8s and Docker in the same host is discouraged, which enables this feature due to microk8s current version is 17.03 and the ARG before FROM feature appeared in 17.05).

ci/package.sh
isutton: I'll keep this one, since having the full name of the binary it holds is a good thing for discovery. Will handle the issues in the internal CI once this work has been merged.

Makefile
```diff
@@ -0,0 +1,30 @@
...
+
+shipper-state-metrics:
+	GOARCH=amd64 CGO_ENABLED=0 GOOS=linux go build -o shipper-state-metrics ./cmd/shipper-state-metrics/*.go
+	docker build -f Dockerfile.shipper-state-metrics -t localhost:32000/shipper-state-metrics:latest --build-arg HTTP_PROXY=$(HTTP_PROXY) --build-arg HTTPS_PROXY=$(HTTPS_PROXY) .
+	docker push localhost:32000/shipper-state-metrics:latest
+
+restart:
+	# Delete all Pods in namespace, to force the ReplicaSet to spawn new ones
+	# with the new latest image (assuming that imagePullPolicy is set to Always).
+	$(KUBECTL) delete po --all
```
icanhazbroccoli: I suggest using full resource type names as it makes the commands to be slightly more readable

pkg/controller/webhook/webhook_controller.go
```diff
@@ -0,0 +1,170 @@
...
+	"net/http"
+
+	"github.com/golang/glog"
+
+	admission_v1beta1 "k8s.io/api/admission/v1beta1"
+	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+
+	shipper_v1alpha1 "github.com/bookingcom/shipper/pkg/apis/shipper/v1alpha1"
```
icanhazbroccoli: Shall we keep it just "shipper"?

parhamdoustdar:
@ksurent, we'll change the CI pipelines after this is merged and Shipper is unblocked. If you don't see any more problems, could you please approve?

pkg/controller/webhook/webhook_controller.go
```diff
@@ -0,0 +1,170 @@
...
+		}
+	}()
+
+	glog.V(2).Info("Started WebHook controller")
+
+	<-stopCh
+
+	glog.V(2).Info("Shutting down WebHook controller")
+
+	if err := server.Shutdown(context.Background()); err != nil {
```
ksurent: This will trigger the `Fatalf` on line 60.

pkg/controller/webhook/webhook_controller.go
```diff
@@ -0,0 +1,170 @@
+package webhook
```
ksurent: This is not a controller and I don't think it belongs under `pkg/controller`. Just `pkg/webhook` should be fine.

Makefile
```diff
@@ -0,0 +1,30 @@
+SHIPPER_NAMESPACE = shipper-system
+KUBECTL = kubectl -n $(SHIPPER_NAMESPACE)
+
+.PHONY: shipper
+
+shipper:
+	GOARCH=amd64 CGO_ENABLED=0 GOOS=linux go build -o shipper ./cmd/shipper/*.go
+	docker build -f Dockerfile.shipper -t localhost:32000/shipper:latest --build-arg HTTP_PROXY=$(HTTP_PROXY) --build-arg HTTPS_PROXY=$(HTTPS_PROXY) .
+	docker push localhost:32000/shipper:latest
```
ksurent: Same question.

Makefile
```diff
@@ -0,0 +1,30 @@
...
+
+shipper-state-metrics:
+	GOARCH=amd64 CGO_ENABLED=0 GOOS=linux go build -o shipper-state-metrics ./cmd/shipper-state-metrics/*.go
+	docker build -f Dockerfile.shipper-state-metrics -t localhost:32000/shipper-state-metrics:latest --build-arg HTTP_PROXY=$(HTTP_PROXY) --build-arg HTTPS_PROXY=$(HTTPS_PROXY) .
+	docker push localhost:32000/shipper-state-metrics:latest
```
ksurent: Same question.

pkg/controller/webhook/webhook_controller.go
```diff
@@ -0,0 +1,170 @@
...
+
+	if err != nil {
+		return &admission_v1beta1.AdmissionResponse{
+			Result: &meta_v1.Status{
+				Message: err.Error(),
+			},
+		}
+	}
+
+	return &admission_v1beta1.AdmissionResponse{
```
ksurent: I don't understand this. The only thing that happens here is we make sure objects are syntactically correct. But the apiserver does that already.
Are we missing a chunk of functionality here? Like, this is where we should be checking for rollout blocks and such.

Makefile
```diff
@@ -0,0 +1,30 @@
+SHIPPER_NAMESPACE = shipper-system
+KUBECTL = kubectl -n $(SHIPPER_NAMESPACE)
+
+.PHONY: shipper
+
+shipper:
+	GOARCH=amd64 CGO_ENABLED=0 GOOS=linux go build -o shipper ./cmd/shipper/*.go
+	docker build -f Dockerfile.shipper -t localhost:32000/shipper:latest --build-arg HTTP_PROXY=$(HTTP_PROXY) --build-arg HTTPS_PROXY=$(HTTPS_PROXY) .
```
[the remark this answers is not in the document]
parhamdoustdar: Microk8s runs a local image registry here. This is a way of building an image from the local directory and push it in a way that will be accessible from the deployment of Shipper running in a microk8s cluster.

kubernetes/shipper.deployment.yaml
```diff
@@ -0,0 +1,37 @@
...
+    matchLabels:
+      app: shipper
+  template:
+    metadata:
+      labels:
+        app: shipper
+    spec:
+      containers:
+      - name: shipper
+        image: localhost:32000/shipper:latest
```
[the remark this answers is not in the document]
parhamdoustdar: This refers to the local microk8s image registry.

pkg/controller/webhook/webhook_controller.go
parhamdoustdar: Right now, our internal version of Kubernetes doesn't check this. So if someone were to create an Application object with the wrong indentation, what happens now is that the Application ends up in the queue, our code fails to decode it, and it ends up blocking the queue until it's fixed.

You're right though; this is also where we will put the logic to take action later, like roll out blocks.

Makefile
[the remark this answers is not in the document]
ksurent: So what do people who prefer minikube do?

pkg/controller/webhook/webhook_controller.go
parhamdoustdar: Fixed.

pkg/controller/webhook/webhook_controller.go
parhamdoustdar: Fixed.

pkg/controller/webhook/webhook_controller.go
```diff
@@ -0,0 +1,170 @@
...
+		}
+
+		if len(body) == 0 {
+			http.Error(w, "empty body", http.StatusBadRequest)
+			return
+		}
+
+		contentType := r.Header.Get("Content-Type")
+
+		if contentType != "application/json" {
```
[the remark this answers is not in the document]
parhamdoustdar: Fixed.

Makefile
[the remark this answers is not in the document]
parhamdoustdar: I made this customizable by using an environment variable in the Makefile. People prefering minikube still have to think of having an image repository, but once that's done, they can just pass it as a variable called `SHIPPER_IMAGE` and it's used when tagging and pushing images.

kubernetes/validating-webhook-configuration.yaml
```diff
@@ -0,0 +1,29 @@
...
+    apiGroups:
+    - "shipper.booking.com"
+    apiVersions:
+    - v1alpha1
+    resources:
+    - applications
+    - releases
+    - installationtargets
+    - capacitytargets
+    - traffictargets
```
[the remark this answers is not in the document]
parhamdoustdar: Awesome catch, thanks. Fixed.

pkg/webhook/webhook.go
```diff
@@ -0,0 +1,176 @@
...
+	"net/http"
+
+	"github.com/golang/glog"
+
+	admission_v1beta1 "k8s.io/api/admission/v1beta1"
+	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+
+	"mime"
```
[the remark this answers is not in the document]
parhamdoustdar: Done!

pkg/webhook/webhook.go
```diff
@@ -0,0 +1,176 @@
...
+		if admissionResponse != nil {
+			admissionReview.Response = admissionResponse
+			if ar.Request != nil {
+				admissionReview.Response.UID = ar.Request.UID
+			}
+		}
+
+		resp, err := json.Marshal(admissionReview)
+		if err != nil {
+			http.Error(w, fmt.Sprintf("could not encode response: %v", err), http.StatusInternalServerError)
```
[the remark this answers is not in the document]
parhamdoustdar: Great catch, fixed. Thanks to you, I realized that there was also another block like this with a missing return statement, and fixed that too.

PR #70, opened 2019-02-21
Implement Release controller as a sub for strategy and schedule controllers

icanhazbroccoli:
A reference to the RFC: https://github.com/bookingcom/shipper/pull/37

This PR introduces a new controller to deal with Release objects. Effectively, it's a combination of Schedule and Strategy controllers.


pkg/controller/release/scheduler_test.go
```diff
@@ -84,7 +174,6 @@ func newScheduler(
 	trafficTargetLister := informerFactory.Shipper().V1alpha1().TrafficTargets().Lister()
 
 	c := NewScheduler(
```
isutton: `release` is not used anywhere anymore, can be removed from `newScheduler`.

pkg/controller/release/scheduler_test.go
```diff
@@ -327,17 +498,19 @@ func TestCreateAssociatedObjectsDuplicateInstallationTargetSameOwner(t *testing.
...
-	if err := c.scheduleRelease(); err != nil {
+	if _, err := c.ScheduleRelease(release.DeepCopy()); err != nil {
 		t.Fatal(err)
 	}
 
 	filteredActions := filterActions(
 		clientset.Actions(),
 		[]string{"update", "create"},
-		[]string{"releases", "installationtargets", "traffictargets", "capacitytargets"},
+		[]string{"installationtargets", "traffictargets", "capacitytargets"},
```
isutton: Don't we have any updates on the Release object we should check after all target objects have been created?

pkg/controller/release/scheduler_test.go
```diff
@@ -514,13 +693,13 @@ func TestCreateAssociatedObjectsDuplicateCapacityTargetNoOwner(t *testing.T) {
...
-	err := c.scheduleRelease()
 
+	_, err := c.ScheduleRelease(release.DeepCopy())
 	if err == nil {
 		t.Fatalf("Expected an error here, none received")
 	}
 
-	if !errors.IsAlreadyExists(err) {
+	if !errors.IsConflict(err) {
 		t.Fatalf("Expected an already-exists error, got: %s", err)
```
isutton: Wrong message.

pkg/controller/release/errors.go
```diff
@@ -220,3 +215,36 @@ func NewDuplicateCapabilityRequirementError(capability string) DuplicateCapabili
 		capability: capability,
 	}
 }
+
+type NotWorkingOnStrategyError error
```
isutton: This doesn't work as expected. You should not use the `error` interface in there, but a concrete type.

The following program prints `this shouldn't happen` when executed:

```go
package main

import "fmt"

type ErrorA error

func IsErrorA(err error) bool {
	_, ok := err.(ErrorA)
	return ok
}

type ErrorB error

func IsErrorB(err error) bool {
	_, ok := err.(ErrorB)
	return ok
}

func main() {
	var err ErrorA = fmt.Errorf("an error")

	if IsErrorB(err) {
		fmt.Printf("this shouldn't happen")
	}
}
```

In contrast, the following program does the right thing, since the custom type is based on a concrete type, not an interface:

```go
package main

import "fmt"

type ErrorA string

func (e ErrorA) Error() string {
	return string(e)
}

func IsErrorA(err error) bool {
	_, ok := err.(ErrorA)
	return ok
}

type ErrorB string

func (e ErrorB) Error() string {
	return string(e)
}

func IsErrorB(err error) bool {
	_, ok := err.(ErrorB)
	return ok
}

func main() {
	var err ErrorA = "an error"

	if IsErrorB(err) {
		fmt.Printf("this shouldn't happen")
	} else {
		fmt.Printf("this is correct")
	}
}
```

pkg/controller/release/release_controller.go
```diff
@@ -0,0 +1,659 @@
...
+		go wait.Until(c.runReleaseWorker, time.Second, stopCh)
+		go wait.Until(c.runApplicationWorker, time.Second, stopCh)
+	}
+
+	glog.V(4).Info("Started Release controller")
+
+	<-stopCh
+}
+
+func (c *Controller) runReleaseWorker() {
```
isutton: Perhaps release and application workers and related code should be placed in distinct files?

pkg/controller/release/release_controller.go
```diff
@@ -0,0 +1,659 @@
...
+
+	defer c.releaseWorkqueue.Done(obj)
+
+	if _, ok := obj.(string); !ok {
+		c.releaseWorkqueue.Forget(obj)
+		runtime.HandleError(fmt.Errorf("invalid object key (will not retry): %#v", obj))
+	}
+	key := obj.(string)
+
+	if shouldRetry := c.syncReleaseHandler(key); shouldRetry {
```
isutton: `shouldRetry` is not used anywhere in the `if` block.

```suggestion
	if c.syncReleaseHandler(key) {
```

pkg/controller/release/release_controller.go
```diff
@@ -0,0 +1,659 @@
...
+
+func (c *Controller) processNextReleaseWorkItem() bool {
+	obj, shutdown := c.releaseWorkqueue.Get()
+	if shutdown {
+		return false
+	}
+
+	defer c.releaseWorkqueue.Done(obj)
+
+	if _, ok := obj.(string); !ok {
```
isutton:
```suggestion
	if key, ok := obj.(string); !ok {
```
The assignment to `key` a few lines below could be then removed.

pkg/controller/release/release_controller.go
```diff
@@ -0,0 +1,659 @@
...
+		c.releaseWorkqueue.Forget(obj)
+		runtime.HandleError(fmt.Errorf("invalid object key (will not retry): %#v", obj))
```
isutton: Missing `return true` at the end of the block.

pkg/controller/release/release_controller.go
```diff
@@ -0,0 +1,659 @@
...
+}
+
+func (c *Controller) processNextAppWorkItem() bool {
+	obj, shutdown := c.applicationWorkqueue.Get()
+	if shutdown {
+		return false
+	}
+	defer c.applicationWorkqueue.Done(obj)
+
+	if _, ok := obj.(string); !ok {
```
isutton:
```suggestion
	if key, ok := obj.(string); !ok {
```
The assignment to `key` a few lines below could be then removed.

pkg/controller/release/release_controller.go
```diff
@@ -0,0 +1,659 @@
...
+		c.applicationWorkqueue.Forget(obj)
+		runtime.HandleError(fmt.Errorf("invalid object key (will not retry): %#v", obj))
+	}
+	key := obj.(string)
+
+	if shouldRetry := c.syncApplicationHandler(key); shouldRetry {
```
isutton:
```suggestion
	if c.syncApplicationHandler(key) {
```

pkg/controller/release/release_controller.go
```diff
@@ -0,0 +1,659 @@
...
+	if shutdown {
+		return false
+	}
+	defer c.applicationWorkqueue.Done(obj)
+
+	if _, ok := obj.(string); !ok {
+		c.applicationWorkqueue.Forget(obj)
+		runtime.HandleError(fmt.Errorf("invalid object key (will not retry): %#v", obj))
+	}
+	key := obj.(string)
```
isutton:
```suggestion

```

pkg/controller/release/release_controller.go
```diff
@@ -0,0 +1,659 @@
...
+		return false
+	}
+
+	defer c.releaseWorkqueue.Done(obj)
+
+	if _, ok := obj.(string); !ok {
+		c.releaseWorkqueue.Forget(obj)
+		runtime.HandleError(fmt.Errorf("invalid object key (will not retry): %#v", obj))
+	}
+	key := obj.(string)
```
isutton:
```suggestion

```

pkg/controller/release/release_controller.go
```diff
@@ -0,0 +1,659 @@
...
+	namespace, name, err := cache.SplitMetaNamespaceKey(key)
+	if err != nil {
+		runtime.HandleError(fmt.Errorf("invalid object key (will not retry): %q", key))
+		return noRetry
+	}
+
+	rel, err := c.releaseLister.Releases(namespace).Get(name)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			glog.V(3).Infof("Release %q has been deleted", key)
```
isutton:
```suggestion
			glog.V(3).Infof("Release %q not found", key)
```

pkg/controller/release/release_controller.go
```diff
@@ -0,0 +1,659 @@
...
+			return noRetry
+		}
+
+		runtime.HandleError(fmt.Errorf("failed to process release %q (will retry): %s", key, err))
+
+		return retry
+	}
+
+	if releaseutil.IsEmpty(rel) {
```
isutton:
```suggestion
	if releaseutil.HasEmptyEnvironment(rel) {
```

pkg/controller/release/release_controller.go
```diff
@@ -0,0 +1,659 @@
...
+		reason, shouldRetry := classifyError(err)
+		condition := releaseutil.NewReleaseCondition(
+			shipper.ReleaseConditionTypeScheduled,
+			corev1.ConditionFalse,
+			reason,
+			err.Error(),
+		)
+		releaseutil.SetReleaseCondition(&rel.Status, *condition)
+
+		if _, err := c.clientset.ShipperV1alpha1().Releases(namespace).Update(rel); err != nil {
```
isutton:
```suggestion
		if _, updateErr := c.clientset.ShipperV1alpha1().Releases(namespace).Update(rel); err != nil {
			err = updateErr
```

pkg/controller/release/release_controller.go
```diff
@@ -0,0 +1,659 @@
...
+			// always retry failing to write the error out to the Release: we need to communicate this to the user
+			return retry
```
isutton:
```suggestion
			shouldRetry = retry
```

pkg/controller/release/release_controller.go
```diff
@@ -0,0 +1,659 @@
...
+func (c *Controller) syncApplicationHandler(key string) bool {
+	namespace, name, err := cache.SplitMetaNamespaceKey(key)
+	if err != nil {
+		runtime.HandleError(fmt.Errorf("invalid object key (will not retry): %q", key))
+		return noRetry
+	}
+	app, err := c.applicationLister.Applications(namespace).Get(name)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			glog.V(3).Infof("Application %q has been deleted", key)
```
isutton:
```suggestion
			glog.V(3).Infof("Application %q not found", key)
```

pkg/controller/release/release_controller.go
```diff
@@ -0,0 +1,659 @@
...
+	}
+
+	if contender == nil {
+		return nil, nil, fmt.Errorf("couldn't find a contender for Application %q",
+			shippercontroller.MetaKey(app))
+	}
+
+	var incumbent *shipper.Release
+	// Walk backwards until we find an installed release that isn't the head of
+	// history. Ffor releases A -> B -> C, if B was never finished this allows C to
```
isutton:
```suggestion
	// history. For releases A -> B -> C, if B was never finished this allows C to
```

pkg/controller/release/scheduler_test.go
icanhazbroccoli: Indeed, this was missing from my original commits, fixed in the newer version.

pkg/controller/release/release_controller.go
icanhazbroccoli: shouldRetry var is there mostly to emphasise the meaning of the return value. In the suggested edit if ```c.syncReleaseHandler(key) {...}``` looks like a successful return code, whereas it's the opposite: shouldRetry = true means something went wrong and the operation should be repeated. I assume, this is my way to say: it's not clear what sync(One|ReleaseHandler|.?) returns.

pkg/controller/release/errors.go
icanhazbroccoli: I removed the type checker function from errors.go as it was never used. What's left is an error definition and a stringer helper, which is there to communicate the error, not to make any decisions based on the specific type.

pkg/controller/release/release_controller.go
isutton: This should be kept to make clear the return value's meaning in the context.

pkg/controller/release/errors.go
isutton: Removing the checker function is fine, but perhaps we should still change the underlying type to at least a string, and add the Error() method to implement the error interface. If a checker function is required in the future, then it won't work as expected and will be a hidden bug.

pkg/controller/release/release_controller.go
```diff
@@ -488,6 +514,8 @@ func (c *Controller) getAssociatedApplicationName(rel *shipper.Release) (string,
 	return appref.Name, nil
 }
 
+// getAssociatedApplicationKey returns an application key in the format:
+// <namespace> / <application name>
```
isutton: I would take the spaces between namespace and application name in there literally. Changing to `"<namespace>/<application name>"` would be more explicit regarding the key's format.

pkg/controller/release/scheduler_test.go
```diff
@@ -332,6 +462,8 @@ func TestCreateAssociatedObjectsDuplicateInstallationTargetMismatchingClusters(t
 	}
 
 	it, tt, ct := buildAssociatedObjects(expected.DeepCopy(), []*shipper.Cluster{cluster.DeepCopy()})
+	// installationtarget already exists, expect an update ection. The rest
```
isutton: `action`

pkg/controller/release/scheduler_test.go
```diff
@@ -654,8 +841,20 @@ func TestCreateAssociatedObjectsDuplicateCapacityTargetSameOwner(t *testing.T) {
 	expected.Status.Conditions = []shipper.ReleaseCondition{
 		{Type: shipper.ReleaseConditionTypeScheduled, Status: corev1.ConditionTrue},
 	}
-	expectedActions := buildExpectedActions(expected.DeepCopy(), []*shipper.Cluster{cluster})
-	expectedActions = filterActions(expectedActions, []string{"update", "create"}, []string{"installationtargets", "traffictargets"})
+
+	it, tt, _ := buildAssociatedObjects(expected.DeepCopy(), []*shipper.Cluster{cluster.DeepCopy()})
+	// 2 create actions: installationtarget and capacitytarget
```
isutton: `traffictarget`

Issue #74, opened 2019-03-19
Store the Shipper deployment object in our repository


parhamdoustdar:
Currently, Shipper's deployment object is a release artefact that we're keeping in Github.

- We should bring it into our repository, so that it's easy to modify, and takes no extra toil
- We should update the documentation to point to this new URL

The first item is done in pull request #60, so we should wait for that to be merged first. Once that's done, we can clean it up and update it, then update the documentation to point to it.

juliogreff:
We now have a deployment object that we use in our CI pipeline after #86 got merged, so now we just need to point our documentation to it.

juliogreff:
Fixed by #182

PR #75, opened 2019-03-19
Add webhook configuration and secret generation to shipperctl

parhamdoustdar:
This pull request is to make it easier to set up the validating web hook introduced in #60. It adds the following functionality to `shipperctl`:

- Create a tls CSR and submit it to Kubernetes as a CertificateSigningRequest object, approve it, and wait until Kubernetes signs the certificate
- Fetch that certificate and put it in a secret
- Create a service to point to the validating webhook

After both this pull request and #60 are merged, we should modify the Deployment object to use the correct name  for the secret(that one is tracked in #74).


cmd/shipperctl/cmd/apply.go
```diff
@@ -120,6 +128,18 @@ func setupManagementCluster(managementCluster *config.ClusterConfiguration, cmd
 		return err
 	}
 
+	if err := createValidatingWebhookSecret(cmd, configurator); err != nil {
+		return err
```
icanhazbroccoli: Do you think it might make sense to decorate the errors (this and 2 more below)? like: ```return fmt.Errorf("failed to create validating webhook secret: %s", err)```

cmd/shipperctl/cmd/apply.go
```diff
@@ -234,6 +254,104 @@ func createManagementServiceAccount(cmd *cobra.Command, configurator *configurat
 	return nil
 }
 
+func createValidatingWebhookSecret(cmd *cobra.Command, configurator *configurator.Cluster) error {
+	fmt.Printf("Checking if a secret already exists for the validating webhook in the %s namespace... ", shipperSystemNamespace)
```
icanhazbroccoli: We have a mix of ```fmt.Printf()``` and ```cmd.Println``` calls, which looks a bit misleading. Is there a way to use a single interface?

cmd/shipperctl/configurator/cluster.go
```diff
@@ -280,3 +294,175 @@ func loadKubeConfig(context, kubeConfigFile string) (*rest.Config, error) {
...
+func (c *Cluster) ApproveShipperCSR() error {
+	csr, err := c.KubeClient.CertificatesV1beta1().CertificateSigningRequests().Get(shipperCSRName, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	approvedCondition := certificatesv1beta1.CertificateSigningRequestCondition{
+		Type:    certificatesv1beta1.CertificateApproved,
+		Reason:  "ShipperctlApprove",
+		Message: "Automatically approved by shipperctl",
```
icanhazbroccoli: Not sure if this makes sense, but would it be helpful if we include the timestamp when it was approved?

cmd/shipperctl/configurator/cluster.go
```diff
@@ -280,3 +294,175 @@ func loadKubeConfig(context, kubeConfigFile string) (*rest.Config, error) {
...
+	return err
+}
+
+// FetchCertificateFromCSR continually fetches the Shipper CSR until
+// it is populated with a certificate and then returns the PEM-encoded
+// certificate from the Status. This is a blocking function.
+//
+// Note that the returned certificate is already PEM-encoded.
+func (c *Cluster) FetchCertificateFromCSR() ([]byte, error) {
+	for retries := 0; retries < 10; retries++ {
```
icanhazbroccoli: Suggestion: make the # of retries a constant

cmd/shipperctl/configurator/cluster.go
```diff
@@ -280,3 +294,175 @@ func loadKubeConfig(context, kubeConfigFile string) (*rest.Config, error) {
...
+		csr, err := c.KubeClient.CertificatesV1beta1().CertificateSigningRequests().Get(shipperCSRName, metav1.GetOptions{})
+		if err != nil {
+			return nil, err
+		}
+
+		if len(csr.Status.Certificate) == 0 {
+			// Pause to give the server some time to sign and populate the certificate
+			time.Sleep(1 * time.Second)
```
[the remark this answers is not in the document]
parhamdoustdar: I think in a long-running process running in the background it would make sense, but since this is a command-line application with no spinners, I'd think that failing fast is better than hanging with no feedback.

What do you think?

cmd/shipperctl/configurator/cluster.go
parhamdoustdar: That one is automatically added by Kubernetes. I ran `shipperctl` locally, and here is what I see in the `conditions` of the CertificateSigningRequest object:

```
  conditions:
  - lastUpdateTime: "2019-04-05T13:35:25Z"
    message: Automatically approved by shipperctl
    reason: ShipperctlApprove
    type: Approved
```

cmd/shipperctl/cmd/apply.go
parhamdoustdar: Very nice catch. Fixed that.

cmd/shipperctl/cmd/apply.go
parhamdoustdar: Do you mean that we should do this for more visibility on the user's side? In that case, I think it's already being done.

An example of output from `shipperctl` first:

```
Setting up management cluster microk8s:
Registering or updating custom resource definitions... done
Creating a namespace called shipper-system... done
(...snipped...)
Creating a secret for the validating webhook:
    Generating a private key... done
    Creating a TLS certificate signing request... done
    Creating a Kubernetes CertificateSigningRequest... done
    Approving the CertificateSigningRequest... error! unable to create new content in namespace shipper-system because it is being terminated
```

Does this do what you're suggesting?

cmd/shipperctl/configurator/cluster.go
parhamdoustdar: Nice suggestion! Done!

cmd/shipperctl/configurator/cluster.go
icanhazbroccoli: Good point, makes sense.

cmd/shipperctl/configurator/cluster.go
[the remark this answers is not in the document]
icanhazbroccoli: Not critical indeed. Let's skip.

cmd/shipperctl/cmd/apply.go
icanhazbroccoli: It does, yes. I was mostly concerned about straight error forwarding which sometimes looses the context of it's occurrence. Doesn't seem to the case here.

PR #83, opened 2019-03-26
clusterclientstore: keep clients per controller per cluster

### LINKED ISSUE #14: Change clusterclientstore to keep clients per controller per cluster
`clusterclientstore` keeps distinct clients per target cluster. We think it makes sense to change that to per target cluster per controller, so that that client throughput is not shared across multiple controllers talking to the same target cluster (think traffic controller getting throttled because installation controller is actively retrying something).

juliogreff:
Closes #14

### DIFF
--- a/pkg/clusterclientstore/cache/cluster.go
+++ b/pkg/clusterclientstore/cache/cluster.go
@@ -18,22 +18,26 @@ const (
 func NewCluster(
 	name,
 	checksum string,
-	client kubernetes.Interface,
 	config *rest.Config,
 	informerFactory kubeinformers.SharedInformerFactory,
+	buildClient clientBuilderFunc,
 	cacheSyncCb func(),
 ) *cluster {
 	return &cluster{
 		state:           StateNotReady,
 		name:            name,
 		checksum:        checksum,
-		client:          client,
 		config:          config,
 		informerFactory: informerFactory,
+		buildClient:     buildClient,
 		cacheSyncCb:     cacheSyncCb,
 		stopCh:          make(chan struct{}),
+
+		clients: make(map[string]kubernetes.Interface),
 	}
 }
+
+type clientBuilderFunc func(string, string, *rest.Config) (kubernetes.Interface, error)
 
 type cluster struct {
 	name string
@@ -41,11 +45,14 @@ type cluster struct {
 	stateMut sync.RWMutex
 	state    string
 
+	clientsMut sync.RWMutex
+	clients    map[string]kubernetes.Interface
+
 	// These are all read-only after initialization, so no lock needed.
 	checksum        string
-	client          kubernetes.Interface
 	config          *rest.Config
 	informerFactory kubeinformers.SharedInformerFactory
+	buildClient     clientBuilderFunc
 
 	cacheSyncCb func()
 
@@ -58,12 +65,26 @@ func (c *cluster) IsReady() bool {
 	return c.state == StateReady
 }
 
-func (c *cluster) GetClient() (kubernetes.Interface, error) {
+func (c *cluster) GetClient(ua string) (kubernetes.Interface, error) {
 	if !c.IsReady() {
-		return c.client, ErrClusterNotReady
+		return nil, ErrClusterNotReady
 	}
 
-	return c.client, nil
+	c.clientsMut.Lock()
+	defer c.clientsMut.Unlock()
+
+	if client, ok := c.clients[ua]; ok {
+		return client, nil
+	}
+
+	client, err := c.buildClient(c.name, ua, c.config)
+	if err != nil {
+		return nil, err
+	}
+
+	c.clients[ua] = client
+
+	return client, nil
 }
 
 func (c *cluster) GetConfig() (*rest.Config, error) {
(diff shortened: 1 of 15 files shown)


pkg/clusterclientstore/cache/cluster.go
```diff
@@ -18,34 +18,41 @@ const (
...
+type clientBuilderFunc func(string, string, *rest.Config) (kubernetes.Interface, error)
+
 type cluster struct {
 	name string
 
 	stateMut sync.RWMutex
 	state    string
 
+	clientsMut sync.RWMutex
+	clients    map[string]kubernetes.Interface
```
icanhazbroccoli: Clients might be sync.Map, in this case clientsMut would be gone.

pkg/clusterclientstore/cache/cluster.go
```diff
@@ -58,12 +65,26 @@ func (c *cluster) IsReady() bool {
 	return c.state == StateReady
 }
 
-func (c *cluster) GetClient() (kubernetes.Interface, error) {
+func (c *cluster) GetClient(ua string) (kubernetes.Interface, error) {
```
icanhazbroccoli: It might be worth leaving a comment what ua means and what it's expected/valid value.

pkg/clusterclientstore/interface.go
```diff
@@ -7,13 +7,13 @@ import (
 )
 
 type ClientProvider interface {
-	GetClient(clusterName string) (kubernetes.Interface, error)
-	GetConfig(clusterName string) (*rest.Config, error)
+	GetClient(string, string) (kubernetes.Interface, error)
```
icanhazbroccoli: I would suggest keeping the variable names in the interface method signatures as this gives an idea what these variables are without checking out a specific implementation.

pkg/clusterclientstore/interface.go
```diff
@@ -7,13 +7,13 @@ import (
...
+	GetConfig(string) (*rest.Config, error)
 }
 
 type Interface interface {
 	AddSubscriptionCallback(SubscriptionRegisterFunc)
 	AddEventHandlerCallback(EventHandlerRegisterFunc)
-	GetClient(string) (kubernetes.Interface, error)
+	GetClient(string, string) (kubernetes.Interface, error)
```
icanhazbroccoli: Same as above: providing argument names might improve the readability.

pkg/clusterclientstore/store.go
```diff
@@ -286,14 +283,22 @@ func (s *Store) create(cluster *shipper.Cluster, secret *corev1.Secret) error {
...
-		}
-	})
+	newCachedCluster := cache.NewCluster(
+		clusterName,
+		checksum,
+		config,
+		informerFactory,
+		func(cluster, ua string, config *rest.Config) (kubernetes.Interface, error) {
+			return s.buildClient(cluster, ua, config)
+		},
```
icanhazbroccoli: I believe you can avoid this anonymous function definition and keep s.buildClient (without parentheses)(please let me know if I'm wrong).

pkg/clusterclientstore/cache/cluster.go
juliogreff: I didn't know about sync.Map, thanks for the suggestion. If I read the docs correctly, though, I could only not *store* a new client if it's already there, but there is no way to only *generate* a client if there isn't one. It's a bit nitpicky though. What do you think?

pkg/clusterclientstore/store.go
juliogreff: I did try :) It complains with the following:

```
pkg/clusterclientstore/store.go:291:4: cannot use s.buildClient (type ClientBuilderFunc) as type "github.com/bookingcom/shipper/pkg/clusterclientstore/cache".clientBuilderFunc in argument to "github.com/bookingcom/shipper/pkg/clusterclientstore/cache".NewCluster
```

I could use the `store.ClientBuilderFunc` type in the `cache` package, but the cyclical dependency doesn't sound right. Open to suggestions here.

pkg/clusterclientstore/cache/cluster.go
icanhazbroccoli: I see, you are right, I must have missed the point. In this case, using a mutex makes more sense. A quick suggestion: consider using sync.Mutex instead as it seems you have no one-writer-multiple-readers kind of pattern, and the aforementioned structure is much more lightweight and faster. Another way to solve this is to use sync.Once(func()), which can guarantee a single-time initialization. All options are quite equivalent in terms of following the best practices. Feel free to choose any :-)

pkg/clusterclientstore/store.go
icanhazbroccoli: Oh, shame on me then. Let's keep it this way, not a big deal.

ksurent: You have two distinct named types: `cache.clientBuilderFunc` and `clusterclientstore.ClientBuilderFunc`. If you really want to keep it a part of the public API, you can put it in a separate package. If the type is deemed more internal, you can just make it not named (just put the entire signature inside `Store`).

PR #85, opened 2019-03-28
controllers: don't retry unrecoverable k8s API errors

### LINKED ISSUE #15: Do not retry unrecoverable k8s API errors
It is possible that an `Update` to an object stored in k8s will make it fundamentally invalid: for example, it is larger than the etcd 1mb storage limit. In this case, we should not continue to retry inserting this object. This may manifest as a 422 "unprocessable entity" HTTP error.

All writes to k8s clusters should inspect their errors to decide if it is possible to retry. If it is possible, the sync handler may return an error to the workqueue. If it is not possible, we should not leave the item in the workqueue.

juliogreff:
It is possible that a write to an object in kubernetes can make it
inherently invalid. For instance, exceeding etcd's 1MB storage limit. In
this case, there is no point to retrying over and over again, since
it'll never recover. Also, having unrecoverable retries clogs the
working queues, preventing more useful work from being done, and also
makes unnecessary requests to the k8s API, making us hit our API request
throttling limit.

Here, we're considering that any .Create() or .Update() that results in
an HTTP 422 "Unprocessable Entity" error is unrecoverable. We can check
that by calling kerrors.IsInvalid(err).

Closes #15

### DIFF
--- a/pkg/controller/application/application_controller.go
+++ b/pkg/controller/application/application_controller.go
@@ -244,19 +244,17 @@ func (c *Controller) syncApplication(key string) bool {
 		app.Spec.RevisionHistoryLimit = &max
 	}
 
-	var shouldRetry bool
-
-	if err = c.processApplication(app); err != nil {
-		shouldRetry = true
-		runtime.HandleError(fmt.Errorf("error syncing Application %q (will retry): %s", key, err))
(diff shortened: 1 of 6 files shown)


juliogreff:
This has been superseded by #91.

PR #91, opened 2019-04-16
Introduce structured errors

### LINKED ISSUE #15: Do not retry unrecoverable k8s API errors

### LINKED ISSUE #90: Shipper needs better error handling
Currently, Shipper retries all errors a small number of times, regardless of the nature of the error: _permanent_ errors that have absolutely no chance of ever succeeding (such as an invalid Application) are retried just as many times as a _transient_ error (such as a network blip caused a cluster to be unreachable for a short amount of time). This is suboptimal for permanent errors because they get retried needlessly and eat up resources, and it's probably a bug for transient errors, as they might not get retried enough and drop off the work queues before the issue that caused them is resolved. It is important that this gets fixed before we're able to run shipper without resyncs (#77).

To solve this, we need to:

1. Identify all of the permanent errors (not only the ones identified in #15) and simply stop retrying them at all.

2. Make the remaining errors be retried for longer than current limit of about 11 retries. Possibly we can keep on retrying them indefinitely if we're sure enough that the underlying cause is gonna be resolved "soon".

To make that happen, I guess this is where we finally introduce structured errors 😊

Also, I have been made aware of the fact that `workqueue.RateLimitingInterface.AddRateLimited` blocks for the period of the required delay, so in effect it can deplete shipper's capacity for handling work, so we need to be very careful with retrying operations for longer periods of time.

juliogreff:
This introduces structured errors in shipper. Errors now have enough knowledge to decide by themselves whether they should be retried or broadcasted, instead of having this logic spread around the controllers. In the process, this PR also prevents a bunch of unrecoverable errors from being retried over and over again, making shipper a bit faster too!

Closes #90, closes #15.


pkg/errors/kubeclient.go
```diff
@@ -0,0 +1,72 @@
...
+	return fmt.Sprintf(
+		"failed API call: %s %s: %q", e.verb, e.resource, e.err,
+	)
+}
+
+// ShouldRetry determines if the error generated by an API call will be
+// retried. It follows API conventions stipulated by
+// https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#http-status-codes
+func (e FailedAPICall) ShouldRetry() bool {
+	switch kerrors.ReasonForError(e) {
```
ksurent: Go's switch cases do not fall through (e.g. like in C). What you want here is comma-separated list of values. And a test.

pkg/clusterclientstore/queue.go
```diff
@@ -109,12 +110,19 @@ func processNextWorkItem(wq workqueue.RateLimitingInterface, handler func(string
 
 	if key, ok = obj.(string); !ok {
 		wq.Forget(obj)
-		runtime.HandleError(fmt.Errorf("invalid object key: %#v", obj))
+		runtime.HandleError(fmt.Errorf("invalid object key (will retry: false): %#v", obj))
```
icanhazbroccoli: I'm wondering if this one deserves it's own structured error too?

pkg/clusterclientstore/store.go
```diff
@@ -161,7 +162,8 @@ func (s *Store) syncCluster(name string) error {
 			return nil
 		}
 
-		return err
+		return shippererrors.NewKubeclientGetError("", name, err).
+			WithShipperKind("Cluster")
```
icanhazbroccoli: Looks a bit foreign to the rest of the API

pkg/controller/capacity/capacity_controller.go
```diff
@@ -340,21 +355,23 @@ func (c *Controller) getSadPods(targetDeployment *appsv1.Deployment, clusterStat
 			fmt.Sprintf("there are %d sad pods", sadPodsCount))
 	}
 
-	return sadPods, nil
+	return sadPods, sadPodsCount == 0, nil
```
icanhazbroccoli: Could you please elaborate the need for an extra boolean flag please? A brief look raises the question: doesn't len(sadPods) === sadPodsCount?

pkg/controller/capacity/capacity_controller.go
```diff
@@ -340,21 +355,23 @@ func (c *Controller) getSadPods(targetDeployment *appsv1.Deployment, clusterStat
...
 }
 
 func (c *Controller) getReport(targetDeployment *appsv1.Deployment, clusterStatus *shipper.ClusterCapacityStatus) (*shipper.ClusterCapacityReport, error) {
-	targetClusterInformer, clusterErr := c.clusterClientStore.GetInformerFactory(clusterStatus.Name)
-	if clusterErr != nil {
+	targetClusterInformer, err := c.clusterClientStore.GetInformerFactory(clusterStatus.Name)
```
icanhazbroccoli: <3

pkg/controller/capacity/deployment.go
```diff
@@ -35,33 +36,41 @@ func (c *Controller) processNextDeploymentWorkItem() bool {
...
 
 	var (
-		item deploymentWorkqueueItem
-		ok   bool
+		key deploymentWorkqueueItem
+		ok  bool
 	)
 
-	if item, ok = obj.(deploymentWorkqueueItem); !ok {
+	if key, ok = obj.(deploymentWorkqueueItem); !ok {
```
icanhazbroccoli: Nice rename, totally makes sense

pkg/errors/clusterclientstore.go
```diff
@@ -0,0 +1,80 @@
...
+type ClusterClientBuildError struct {
+	clusterName string
+	err         error
+}
+
+func (e ClusterClientBuildError) Error() string {
+	return fmt.Sprintf("cannot create client for cluster %q: %s", e.clusterName, e.err.Error())
+}
+
+func (e ClusterClientBuildError) ShouldRetry() bool {
```
icanhazbroccoli: It might be worth defining corresponding interfaces and enforce conformation like: `var _ RetriableError = ClusterClientBuildError{}`; same for `ShouldBroadcast`.

icanhazbroccoli: Did not notice the interface definition below, my bad

pkg/clusterclientstore/queue.go
juliogreff: I wouldn't mind turning it into one, but I also don't see the point here. `HandleError` doesn't actually *handle* anything, it just logs it as a string. In fact, it's horrible that we have to do this dance of turning it into an error first just to have it stringified again right afterward.

pkg/clusterclientstore/store.go
juliogreff: I agree, I'm not a fan of it either, but that's the best I could come up with so far. Open to suggestions here.

For reference, I started with the following, but decided against it due to verbosity:

```go
return shippererrors.NewKubeclientGetError(shipper.SchemeGroupVersion.WithKind("Cluster"), "", name, err)
```

pkg/controller/capacity/capacity_controller.go
juliogreff: In one of the early returns, we return a list of `sadPods` that may be empty, but it doesn't necessarily mean that the cluster is ok, as it can be in `conditions.WrongPodCount`. It's indeed a function that deserves some refactoring, as it doesn't just get sad pods.

PR #96, opened 2019-05-15
shipper-state-metrics: collect time spent on pending release conditions

juliogreff:
We're trying to collect metrics on how long rollouts take, end to end.
Although we can't currently get that directly, we have a decent proxy
metric: how long a release has spent with a Strategy.Condition.Status=False.

We will start exposing that through a prometheus summary metric through
shipper-state-metrics, with the 50th, 90th and 99th percentile. I chose
not to go with a histogram because we currently have no idea about the
distribution of these numbers, so it's really hard to set up buckets
that make sense. Also, I don't expect many aggregations to be done on
this, as we only have one instance of shipper-state-metrics running,
collecting data for all avaiable releases.

For getting these numbers back, for instance the 90th percentile we've
been waiting for releases with ContenderAchievedInstallation=False, one
could query prometheus with the following:

	shipper_release_durations{cond_type="ContenderAchievedInstallation", quantile="0.9"}

Currently cond_type="ContenderAchievedInstallation" gives me the wrong impression that a release *has* in fact ContenderAchievedInstallation=True, so I'm considering a different name (perhaps "false_cond"). I also need a better name for shipper_release_durations, but I haven't gotten to a good one yet. Suggestions very much welcome.

### DIFF
--- a/cmd/shipper-state-metrics/math.go
+++ b/cmd/shipper-state-metrics/math.go
@@ -1,0 +1,68 @@
+package main
+
+import (
+	"errors"
+	"math"
+	"sort"
+)
+
+// Copyright (c) 2014-2015 Montana Flynn (https://anonfunction.com)
+// Licensed under the MIT License (check NOTICE for full license)
+// Obtained from https://github.com/montanaflynn/stats, minor modifications
+// were applied
+func Percentile(input []float64, percent float64) (percentile float64, err error) {
+	// Find the length of items in the slice
+	il := len(input)
+
+	// Return an error for empty slices
+	if il == 0 {
+		return 0, errors.New("empty input")
+	}
+
+	// Return error for less than 0 or greater than 100 percentages
+	if percent < 0 || percent > 100 {
+		return 0, errors.New("percent out of bounds")
+	}
+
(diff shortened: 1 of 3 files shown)


cmd/shipper-state-metrics/math.go
```diff
@@ -0,0 +1,65 @@
+package main
+
+import (
+	"errors"
+	"math"
+	"sort"
+)
+
+// Lifted from https://github.com/montanaflynn/stats
```
ksurent: This should probably be mentioned in NOTICE as well? I don't really know but worth checking.

cmd/shipper-state-metrics/math.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+func MakeSummary(input []float64, quantiles []float64) map[float64]float64 {
+	summary := make(map[float64]float64)
+
+	for _, p := range quantiles {
+		percentile, err := Percentile(input, p)
+		summary[p] = percentile
+		if err != nil {
+			panic(err)
```
icanhazbroccoli: Feels like we might prefer returning another error rather than panic here.

PR #97, opened 2019-05-15
Release 0.2: patched CapacityController preventing etcd from running out of space

icanhazbroccoli:


icanhazbroccoli:
This MR was opened with a wrong action intention: the merge was not supposed to reach master branch directly but to release-0.2. It would be back-ported to master eventually. The upcoming releases are guaranteed to contain the fix.

PR #101, opened 2019-06-06
RFC: Replace Helm charts with Kustomize for application packaging

juliogreff:
This is a proposal for how we'd use kustomize in Shipper, if we were to ditch helm charts. What I want to get out of this RFC is:

1. With a more concrete suggestion of how things could look like (this RFC), do we wanna move away from helm into kustomize?

2. Assuming we do, is the proposed way the actual way we wanna go? Did I miss any challenges? Will this make users' lives more difficult or easier? Does it actually solve the problems it's supposed to solve? Does it create any new ones I didn't foresee?

### DIFF
--- a/rfcs/999-kustomize.adoc
+++ b/rfcs/999-kustomize.adoc
@@ -1,0 +1,209 @@
+= 999 Replace Helm charts with Kustomize for application packaging
+Julio Greff <julio.deoliveira@booking.com>
+:RFC-Status: Draft
+
+== Motivation
+
+Shipper currently uses link:https://helm.sh/[Helm] charts to package
+applications. Although it has been working fine so far, several things motivate
+us to look for an alternative way of packaging applications.
+
+First, link:https://sweetcode.io/a-first-look-at-the-helm-3-plan/[Helm 3], the
+upcoming version of Helm, will allow Lua scripting inside charts. We believe
+it's not appropriate to have a scripting engine running arbitrary code inside a
+server side component such as Shipper, as it would be relatively easy for a
+misbehaved chart to pin down Shipper workers, potentially denying service to
+other well behaved applications.
+
+Second, Helm generates Kubernetes objects by templating YAML files using Go
+templates. Generating well formed YAML files this way is very error prone, as
+one needs to pay attention to correctly balanced quotes, keeping track of
+indentation levels and others. Helm's choice of using Go templates also
+prevents us from instrospecting charts with languages other than Go.
+
+Lastly, Helm charts only allow configuration of values that the chart authors
+have thought of, and wrote templates for. If the need arises to change a value
+that's hardcoded in the chart, or add a completely new one that's not present,
+it's necessary to fork the chart, and add the changes there. We would prefer to
+allow users to apply arbitrary patches to objects instead, as it would reduce,
+or completely remove, the need to fork charts.
+
+link:https://kustomize.io/[Kustomize] brands itself as a Kubernetes native
+configuration management that does not rely on templates to customize
+application configuration. It uses plain YAML artifacts, allowing easier
+instrospection from any programming language. It also allows the application of
+arbitrary patches to any of the resources it manages, allowing more freedom to
+users of off-the-shelf configuration bundles to customize to their heart's
+content.
+
+As Kustomize seems to address the challenges with our current usage of Helm,
+this RFC aims to discuss how a potential implementation of Kustomize would work
+in Shipper.
+
+== Reference level explanation
+
+A common use case for us is to have a Shipper application referring to a
+centrally supported bundle of configuration, with sane defaults for basic
+infrastructure, and a placeholder for the application's Docker image.
+Additionally, applications should be able to enable or disable bigger chunks of
+configuration, such as which optional sidecars (out of either a collection of
+centrally supported sidecars, or completely bespoke ones) should be present.
+Today, that's supported by a Helm chart specified by `.spec.template.chart`,
+and tweaks to the configuration are injected as values through
+`.spec.template.values`, both in the Application object.
+
+Using Kustomize, the Application object would lose both `.spec.template.chart`
+and `.spec.template.values`, having those replaced by a single
+`.spec.template.kustomization`, as exemplified below.
+
+.application.yaml
+[source,yaml]
+----
+apiVersion: shipper.booking.com/v1alpha1
+kind: Application
+metadata:
+  name: bikerental
+spec:
+  template:
+    clusterRequirements:
+      regions:
+      - name: eu-nl
+      capabilities: []
+    kustomization:
+      name: https://example.com/juliogreff/bikerental # <1>
+      image:
+        name: juliogreff/bikerental
+        tag: deadcafe
+    strategy:
+      steps:
+      - name: full on
+        capacity:
+          incumbent: 0
+          contender: 100
+        traffic:
+          incumbent: 0
+          contender: 100
+----
+<1> This example does not take versioning into account. That can be easily
+addressed by pointing to an immutable URL, such as
+`https://raw.githubusercontent.com/juliogreff/bikerental/$GIT_REF/`, where
+`$GIT_REF` is a commit or tag name. This is not a final decision, as versioning
+of kustomizations is not a solved problem yet.
+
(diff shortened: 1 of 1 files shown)


kanatohodets:
Nice, the inline patch looks really good to me. Thinking about use cases: if I want to add a totally new Kubernetes object to my deployment, like `my-cool-configmap`, how would I do it?

Could I use a "role"?
```yaml
apiVersion: shipper.booking.com/v1alpha1
kind: Application
metadata:
  name: bikerental
spec:
  template:
    clusterRequirements: {}
    kustomization:
      image:
        name: juliogreff/bikerental
        tag: deadcafe
      bases:
        - https://example.com/core
        # like this? 
        - https://example.com/my-cool-configmap
      namePrefix: bikerental-
      commonLabels:
        app: bikerental
      patchesStrategicMerge:
        - https://example.com/sidecars/sidecar.yaml
    ....
```

Or would I need an inheritance hierarchy?
```yaml
apiVersion: shipper.booking.com/v1alpha1
kind: Application
metadata:
  name: bikerental
spec:
  template:
    clusterRequirements: {}
    kustomization:
      image:
        name: juliogreff/bikerental
        tag: deadcafe
      bases:
        # hopefully not a fork, but just inherited... 
        # and then I guess something happens with version resolution internally... aaagh
        - https://example.com/core-with-my-cool-configmap 
      namePrefix: bikerental-
      commonLabels:
        app: bikerental
      patchesStrategicMerge:
        - https://example.com/sidecars/sidecar.yaml
     ...
```

If the approach is more like a role, I'm pretty excited about this. I assume so, since the field is a list.

Some things we'll need to figure out, I think:

* version resolution + caching strategy for kustomize bases/patches in shipper.
* (internally, but I also think this is a problem in the wider ecosystem) hosting kustomize patches somewhere shipper can pull them from, ideally with versioning/access control/etc. 

Regarding versioning of the bases/patches: a key requirement is the ability for Shipper to request a 'floating' version like 0.3.x, but resolve that to a concrete version, like 0.3.1. This is akin to Docker tags + image SHA1s. Just like our plan for floating Chart version ranges, `Release` objects should always have a concrete version in them, so that rollbacks can happen without needing the Kustomize repository available. This might be a good question to raise on `#kustomize` on the K8s slack.

rfcs/999-kustomize.adoc
```diff
@@ -0,0 +1,197 @@
...
+metadata:
+  name: bikerental
+spec:
+  template:
+    clusterRequirements:
+      regions:
+      - name: eu-nl
+      capabilities: []
+    kustomization:
+      image:
```
kanatohodets: How does Kustomize know which object in the bundle to apply this patch to?

rfcs/999-kustomize.adoc
```diff
@@ -0,0 +1,197 @@
...
+        name: juliogreff/bikerental
+        tag: deadcafe
+      bases:
+        - https://example.com/core
+      namePrefix: bikerental-
+      commonLabels:
+        app: bikerental
+      patchesStrategicMerge:
+        - https://example.com/sidecars/sidecar.yaml
```
kanatohodets: The Helm sidecar approach allows the sidecars chart to specify a set of parameters on a per-sidecar basis, which change how the sidecar containers are rendered. For example, the dynamic configuration sidecar lists which pieces of dynamic configuration it should track. How would we pass parameters to sidecars in this strategy? Something around a configmap per sidecar?

rfcs/999-kustomize.adoc
```diff
@@ -0,0 +1,197 @@
...
+----
+<1> Kustomizations can include as many of any Kubernetes resources as they
+need. As those are not of consequence to this RFC, they are omitted for
+brevity.
+
+The kustomizations above cover the application level configuration. At release
+time, Shipper still needs to apply its own little bit of configuration with
+data from the Release object it just created. That kustomization would be the
+equivalent of the following YAML snippet. It injects references to the release
+hash through `nameSuffix` and `commonLabels`, adjusts the amount of replicas
```
kanatohodets: The `commonLabels` functionality is nice. We might be able to lean on this to clean up some of the gnarly object mutation bits in the Installation controller.

rfcs/999-kustomize.adoc
```diff
@@ -0,0 +1,197 @@
...
+images:
+  - name: app
+    newName: juliogreff/bikerental
+    newTag: deadcafe
+----
+
+== Open questions
+
+* The migration path for Shipper to move from Helm charts to Kustomize is
+  entirely unclear. We'll probably need to have both ways of packaging
```
kanatohodets: Yeah, I think dual life is inevitable, at least for a while. It _might_ be possible to automatically convert 'values' sections to patches, but it might not be worth the effort. My first cut would probably be just distinct 'chart' and 'kustomize' fields, a validation in the webhook which prevents both from being populated at once, or 'values' from being populated if 'kustomize' is.

rfcs/999-kustomize.adoc
juliogreff: First of all, my mistake, I got the schema wrong ^_^ It should be `images`. Second of all, it's not a patch, it's a transformer, making it somewhat involved from kustomize's side. [Documentation here](https://github.com/kubernetes-sigs/kustomize/blob/master/docs/fields.md#images).

juliogreff:
@kanatohodets re: adding a completely new object:

> Could I use a "role"?

Yes. The only downside of having multiple bases I found so far is that you can't patch objects in one base from the other.

> Or would I need an inheritance hierarchy?

You can do that too, and you get to patch objects from your base as well as a bonus.

BUT! The idiomatic way of doing it is just by declaring new resources, directly in `.spec.kustomize.resources`. [Docs here](https://github.com/kubernetes-sigs/kustomize/blob/master/docs/fields.md#resources-1).

```yaml
apiVersion: shipper.booking.com/v1alpha1
kind: Application
metadata:
  name: bikerental
spec:
  template:
    clusterRequirements: {}
    kustomization:
      image:
        name: juliogreff/bikerental
        tag: deadcafe
      bases:
        - https://example.com/core
      resources:
        - https://example.com/my-cool-configmap.yaml
     ...
```

kanatohodets:
I think the only question remaining for me is about sidecar configuration. As discussed in person, we'll shift from sidecars being configured directly in their rendered YAML (with `if` conditions in gotmpl, like Helm) to consuming a ConfigMap which the user can work with. So the user would provide a 'patchStrategicMergeInline` which modifies the appropriate ConfigMap. Can Kustomize mutate contents of ConfigMaps in a sensible way?

This also might raise some questions about the ordering of `patchStrategicMergeInline` vs `patchStrategicMerge`.

kanatohodets:
I lied, another question: we have some resources which should be suffixed with a per-release `deadbeef` (most of them) and some which should not (the main `Service` object, which should overwrite whatever is currently in the cluster). I think you mentioned a strategy for this, but I don't remember what it was. How can we handle these cases, where the 'suffix' should not be universally applied?

kanatohodets:
I think the answer to the sidecar ConfigMap configuration question is https://github.com/kubernetes-sigs/kustomize/blob/master/docs/fields.md#configmapgenerator

rfcs/999-kustomize.adoc
```diff
@@ -0,0 +1,265 @@
...
+`chart` and `values` are populated, or only `kustomize`. At some point, after a
+few Shipper versions, the validating webhook can start rejecting objects with
+`chart` and `values`, while Shipper itself will still support both setups for a
+few more versions. Ultimately, Shipper will stop supporting Helm charts
+entirely, and any remaining applications using it will be invalid, and will no
+longer be managed by Shipper until they're fixed.
+
+Since we do Semantic Versioning, and to avoid the need of a new major version
+release when we do remove support for Helm charts, Helm support should be
+dropped before Shipper turns 1.0.
```
isutton: Would be nice to have this migration plan as a series of steps instead of a paragraph to emphasize the scope of each step.

kanatohodets:
It looks like Kustomize supports inline patches [as of September this year](https://github.com/kubernetes-sigs/kustomize/blob/master/docs/v3.2.0.md), which is cool!

I think the answer to my question about suffixing some resources with the hash for that release is that we'd include a custom transformer in our implementation.

PR #105, opened 2019-06-24
Helm chart version range

icanhazbroccoli:
Introduced Helm chart version ranges

This change makes Shipper to be a real Helm repo client: from now on, an index file is being fetched first, and then a version satisfying a provided SemVer constraint will be used. The most up-to-date version would be picked among all satisfying the constraint.

The version range is being resolved only once: at the Application controller cycle. Later on, on Release and Installation controller cycles the version is being picked up from the corresponding object specs.

If a helm repo us unavailable, the shipper will try to lookup for an index file locally in the FS cache. If it exists, it will pick the most up-to-date chart which is available locally. I.e.: if the index file suggests versions A, B and C (increasing order, so the preferred version should be C) to satisfy the input version constraint, and version C is unavailable locally whereas A and B are cached, version B would be chosen as the most up-to-date satisfying the constraint.


pkg/controller/release/scheduler_test.go
```diff
@@ -26,6 +30,16 @@ func init() {
 	releaseutil.ConditionsShouldDiscardTimestamps = true
 }
 
+var localFetchChart = func(chartspec *shipper.Chart) (*helmchart.Chart, error) {
+	data, err := ioutil.ReadFile(path.Join("testdata", "simple-0.0.1.tgz"))
```
juliogreff: Is this chart name supposed to be hardcoded? I'd expect it to be formed from them chartspec.

pkg/controller/release/testmain_test.go
```diff
@@ -1,28 +0,0 @@
...
-	"os"
-	"testing"
-
-	"k8s.io/helm/pkg/repo/repotest"
-)
-
-var chartRepoURL string
-
-func TestMain(m *testing.M) {
-	srv, hh, err := repotest.NewTempServer("testdata/*.tgz")
```
juliogreff: I did not even know we did this, but makes me happy to see it gone <3

pkg/controller/installation/installer_test.go
```diff
@@ -16,13 +20,30 @@ import (
...
 	shippertesting "github.com/bookingcom/shipper/pkg/testing"
 )
 
+var localFetchChart = func(chartspec *shipper.Chart) (*chart.Chart, error) {
+	data, err := ioutil.ReadFile(
+		path.Join(
+			"testdata",
+			"chart-cache",
+			"localhost",
+			fmt.Sprintf("%s-%s.tgz", chartspec.Name, chartspec.Version),
```
juliogreff: I'm of two minds with hardcoding localhost here. On one hand, it may never even matter. On the other hand, it doesn't match the actual RepoURL in the chartspec, which may make it for a very frustrating time for someone debugging around here down the road.

pkg/chart/repo/fscache.go
```diff
@@ -0,0 +1,74 @@
...
+
+func (f *fsCache) Fetch(name string) ([]byte, error) {
+	name = clean(name)
+	path := filepath.Join(f.dir, name)
+	return ioutil.ReadFile(path)
+}
+
+func (f *fsCache) Store(name string, data []byte) error {
+	if err := f.ensureLimits(len(data)); err != nil {
+		return nil
```
juliogreff: Shouldn't this return `err`?

pkg/controller/installation/installer.go
```diff
@@ -54,7 +56,7 @@ func NewInstaller(chartFetchFunc shipperchart.FetchFunc,
 // cluster, or an error.
 func (i *Installer) renderManifests(_ *shipper.Cluster) ([]string, error) {
 	rel := i.Release
-	chart, err := i.fetchChart(rel.Spec.Environment.Chart)
+	chart, err := i.chartFetcher(&rel.Spec.Environment.Chart)
```
juliogreff: We may or may not decide to do this in the scope of this PR, but it would be nice if the `chartFetcher` returned structured errors, so we could avoid wrapping everything with a `RenderManifestError`. I would expect some errors to be more retriable than others.

pkg/controller/release/testmain_test.go
icanhazbroccoli: My bad: apparently, it was an easy way to plug it back when I was working on release controller, and never fixed it since then. Clearly there is no place for this kind of hacks in unit testing.

pkg/chart/repo/handlers.go
```diff
@@ -0,0 +1,51 @@
...
+		cv, err := repo.ResolveVersion(chartspec.Name, chartspec.Version)
+		if err != nil {
+			return nil, fmt.Errorf("failed to resolve chart version: %v", err)
+		}
+
+		return cv, nil
+	}
+}
+
+var FetchChartFunc = func(c *Catalog) ChartFetcher {
```
juliogreff: Out of curiosity, why are this and `ResolveChartVersionFunc` assigned to vars instead of being a normal func? It makes me think there's something special to them, but I was not able to figure out why.

pkg/chart/repo/repo.go
```diff
@@ -0,0 +1,284 @@
...
+
+	// copy-paste from Helm's chart_downloader.go
+	chartURL, err := url.Parse(cv.URLs[0])
+	if err != nil {
+		return nil, fmt.Errorf("invalid chart URL format: %v", cv.URLs[0])
+	}
+
+	// If the URL is relative (no scheme), prepend the chart repo's base URL
+	if !chartURL.IsAbs() {
+		repoURL, _ := url.Parse(r.url) // URL is validated before
```
juliogreff: I think doing this URL parsing when creating the repo gives a stronger guarantee that it is in fact valid. Otherwise, it's relying on the detail that it's done in `RefreshIndex`, which may or may not hold true forever.

pkg/chart/repo/repo.go
```diff
@@ -0,0 +1,284 @@
...
+	i.SortEntries()
+	if i.APIVersion == "" {
+		// do not support pre-v2.0.0
+		return i, repo.ErrNoAPIVersion
+	}
+
+	return i, nil
+}
+
+// Redefined in tests
```
juliogreff: It looks like both `loadChart` and `loadIndex` aren't redefined in tests.

pkg/chart/repo/repo.go
```diff
@@ -0,0 +1,284 @@
...
+	}
+
+	if len(versions) == 0 {
+		return nil, repo.ErrNoChartVersion
+	}
+
+	var highestver *repo.ChartVersion
+	var lasterr error
+	for _, ver := range versions {
+		if _, lasterr = r.Fetch(ver); lasterr == nil {
```
juliogreff: This sounds to me as always trying to fetch charts and not really using the cache a whole lot. Is that on purpose? If not, I would've expected to see `FetchIfNotCached` here.

pkg/controller/application/application_controller_utils.go
```diff
@@ -30,7 +30,7 @@ func (c *Controller) createReleaseForApplication(app *shipper.Application, relea
 			Labels: map[string]string{
 				shipper.ReleaseLabel:                releaseName,
 				shipper.AppLabel:                    app.Name,
-				shipper.ReleaseEnvironmentHashLabel: hashReleaseEnvironment(app.Spec.Template),
+				shipper.ReleaseEnvironmentHashLabel: hashReleaseEnvironment(app.Spec.Template), // TODO check logic
```
juliogreff: It does look like `hashReleaseEnvironment` still does the right thing, did you have something specific in mind?

pkg/chart/repo/handlers.go
icanhazbroccoli: Good catch! It's an artefact: initially I redefined these functions in the tests, but later on transitioned to module attribute injection. To be fixed.

PR #106, opened 2019-06-24
Rollout Blocks

### LINKED ISSUE #1: Rollout blocks
The system should be able to support per–namespace and global rollout blocks. This interface should be easy to engage and disengage. A block should include a human message describing why the block is in place.

While the block is in place we should not create any new Release objects.

The blocks should be 'soft' so that you can override the block by confirming intent.

hbarkal:
This introduces RolloutBlocks as a CRD in shipper. 
Blocking rollouts locally or globally is possible using a RolloutBlock object in the relevant namespace.

Closes #1

### DIFF
--- a/ci/e2e.sh
+++ b/ci/e2e.sh
@@ -12,7 +12,7 @@ kubectl wait -n container-registry --for=condition=ready pod/$REGISTRY_POD
 
 # Build an image with the test charts and deploy it
 HELM_IMAGE=localhost:32000/bookingcom/shipper-helm:latest make helm
-sed s=\<HELM_IMAGE\>=localhost:32000/bookingcom/shipper:latest= ci/helm.yaml | \
+sed s=\<HELM_IMAGE\>=localhost:32000/bookingcom/shipper-helm:latest= ci/helm.yaml | \
 	kubectl apply -f -
 
 # Build an image with shipper and deploy it


pkg/apis/shipper/v1alpha1/types.go
```diff
@@ -508,6 +511,47 @@ const (
 	StrategyStateFalse   StrategyState = "False"
 )
 
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// A RolloutBlock defines the a state where rollouts are blocked, locally or globally.
+// This is used by the ApplicationController to unable rollouts.
```
icanhazbroccoli: s/unable/disable/

pkg/controller/application/rolloutblock.go
```diff
@@ -0,0 +1,66 @@
...
+	} else if len(overrideRB) > 0 {
+		c.recorder.Event(app, corev1.EventTypeNormal, "Overriding RolloutBlock", overrideRB)
+	}
+	return !overrideRolloutBlock
+}
+
+func (c *Controller) removeRolloutBlockFromAnnotations(overrideRB string, rbName string, app *shipper.Application) {
+	overrideRBs := strings.Split(overrideRB, ",")
+	overrideRBs = stringUtil.Delete(overrideRBs, rbName)
+	app.Annotations[shipper.RolloutBlocksOverrideAnnotation] = strings.Join(overrideRBs, ",")
```
icanhazbroccoli: It might be worth sorting overrideRBs before concatenation in order to make sure we're generating the same results if this cycle steps in once again.

pkg/controller/application/rolloutblock.go
```diff
@@ -0,0 +1,66 @@
...
+	if err != nil {
+		runtime.HandleError(err)
+	}
+}
+
+func (c *Controller) updateApplicationRolloutBlockCondition(rbs []*shipper.RolloutBlock, app *shipper.Application) {
+	if len(rbs) > 0 {
+		var sb strings.Builder
+		for _, rb := range rbs {
+			sb.WriteString(rb.Name + " ")
```
icanhazbroccoli: Here you have a concat char, which would be a trailing symbol in the end. I might suggest initialising a simple slice of `[]string` and appending + strings.Concat in the end. The amount of data is insignificant, which means there would be no visible performance penalty.

icanhazbroccoli: With a slice approach you can get rid of this if-else case for free.

pkg/controller/release/rolloutblock.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	overrideRolloutBlock, eventMessage, err := rolloutblockUtil.ShouldOverrideRolloutBlock(relOverrideRB, nsRBs, gbRBs)
+	if err != nil {
+		switch errT := err.(type) {
+		case shippererrors.InvalidRolloutBlockOverrideError:
+			// remove from annotation!
+			rbName := err.(shippererrors.InvalidRolloutBlockOverrideError).RolloutBlockName
+			s.removeRolloutBlockFromAnnotations(relOverrideRB, rbName, rel)
+			err = nil
```
icanhazbroccoli: Any specific reasons to nullify this error var?

pkg/controller/release/rolloutblock.go
```diff
@@ -0,0 +1,63 @@
...
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/runtime"
+
+	shipper "github.com/bookingcom/shipper/pkg/apis/shipper/v1alpha1"
+	shippererrors "github.com/bookingcom/shipper/pkg/errors"
+	rolloutblockUtil "github.com/bookingcom/shipper/pkg/util/rolloutblock"
+	stringUtil "github.com/bookingcom/shipper/pkg/util/string"
+)
+
+func (s *Scheduler) shouldBlockRollout(rel *shipper.Release) (bool, string, error) {
```
icanhazbroccoli: Not sure if the semantics make sense to me: on the application level rollout block is a responsibility of the controller itself, and in this controller we delegate an equivalent responsibility to scheduler.
I assume, this might be taken as 2 action points:
1. Does it have to be a responsibility of the scheduler? Can it be moved to the controller?
2. Application-level and release-level blocks don't look very much different, do you think it's possible to have a single component responsible for it and embeddable into distinct controllers?

pkg/controller/release/rolloutblock.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	nsRBs, err := s.rolloutBlockLister.RolloutBlocks(rel.Namespace).List(labels.Everything())
+	if err != nil {
+		runtime.HandleError(fmt.Errorf("error syncing Release %q Because of namespace RolloutBlocks (will retry): %s", rel.Name, err))
+	}
+
+	gbRBs, err := s.rolloutBlockLister.RolloutBlocks(shipper.ShipperNamespace).List(labels.Everything())
+	if err != nil {
+		runtime.HandleError(fmt.Errorf("error syncing Release %q Because of global RolloutBlocks (will retry): %s", rel.Name, err))
```
icanhazbroccoli: I suggest reflecting the error essentials here, like: `failed to list rollout block objects: %s`

pkg/controller/release/rolloutblock.go
```diff
@@ -0,0 +1,63 @@
...
+			runtime.HandleError(fmt.Errorf("error of type %T overriding rollout block %s", errT, err.Error()))
+			return true, "", err
+		}
+	}
+
+	if overrideRolloutBlock && len(relOverrideRB) > 0 {
+		s.recorder.Event(rel, corev1.EventTypeNormal, "Override RolloutBlock", relOverrideRB)
+	}
+
+	return !overrideRolloutBlock, eventMessage, err
```
icanhazbroccoli: returning `false` explicitly makes this bit more readable

pkg/controller/release/rolloutblock.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+func (s *Scheduler) removeRolloutBlockFromAnnotations(overrideRB string, rbName string, release *shipper.Release) {
+	overrideRBs := strings.Split(overrideRB, ",")
+	overrideRBs = stringUtil.Delete(overrideRBs, rbName)
```
icanhazbroccoli: I'd suggest refactoring it all around and calling this method Grep

pkg/controller/rolloutblock/release.go
```diff
@@ -0,0 +1,68 @@
...
+
+	if rolloutBlock.DeletionTimestamp != nil {
+		return fmt.Errorf("RolloutBlock %s/%s has been deleted", rolloutBlock.Namespace, rolloutBlock.Name)
+	}
+
+	if rolloutBlock.Status.Overrides.Release == nil {
+		return nil
+	}
+
+	rolloutBlock.Status.Overrides.Release = stringUtil.Delete(rolloutBlock.Status.Overrides.Release, relFullName)
```
icanhazbroccoli: Same as above: Grep would make slightly more sense

pkg/controller/rolloutblock/rolloutblock_controller.go
```diff
@@ -0,0 +1,622 @@
...
+
+// NewController returns a new RolloutBlock controller.
+func NewController(
+	shipperClientset clientset.Interface,
+	informerFactory shipperinformers.SharedInformerFactory,
+	recorder record.EventRecorder,
+) *Controller {
+	applicationInformer := informerFactory.Shipper().V1alpha1().Applications()
+	releaseInformer := informerFactory.Shipper().V1alpha1().Releases()
+	rolloutblockInformer := informerFactory.Shipper().V1alpha1().RolloutBlocks()
```
icanhazbroccoli: would it make sense to keep camel case naming and change it to rolloutBlockInformer?

pkg/controller/rolloutblock/rolloutblock_controller.go
```diff
@@ -0,0 +1,622 @@
...
+			rbKey,
+			OverridingRelease,
+			key,
+			false,
+		}
+		c.rolloutblockWorkqueue.Add(rolloutBlockUpdater)
+	}
+}
+
+func (c *Controller) deleteRelease(obj interface{}) {
```
icanhazbroccoli: Not sure the method name reflects on what's happening inside. Should it be rather called like: enqueueRolloutBlock(Updater)?

pkg/controller/rolloutblock/rolloutblock_controller.go
```diff
@@ -0,0 +1,622 @@
...
+			rbKey,
+			OverridingRelease,
+			key,
+			true,
+		}
+		c.rolloutblockWorkqueue.Add(rolloutBlockUpdater)
+	}
+}
+
+func (c *Controller) updateRelease(oldObj, newObj interface{}) {
```
icanhazbroccoli: again: it doesn't really update release

pkg/controller/rolloutblock/rolloutblock_controller.go
icanhazbroccoli: I missed the initial intention of this method. I would suggest something like: onDeleteRelease instead, so it provider an event reactor context.

pkg/controller/rolloutblock/rolloutblock_controller.go
```diff
@@ -0,0 +1,622 @@
...
+	if !oldOk {
+		oldOverride = ""
+	}
+	if !newOk {
+		newOverride = ""
+	}
+
+	newOverrides := strings.Split(newOverride, ",")
+	oldOverrides := strings.Split(oldOverride, ",")
+	if stringUtil.Equal(newOverrides, oldOverrides) {
```
icanhazbroccoli: if we compose them in a deterministic way (which is the case if we stable-sort them), a simple string comparison is enough without parsing and a granular per-element comparison

pkg/controller/rolloutblock/rolloutblock_controller.go
```diff
@@ -0,0 +1,622 @@
...
+	}
+
+	defer c.rolloutblockWorkqueue.Done(obj)
+
+	var (
+		ok                  bool
+		rolloutBlockUpdater RolloutBlockUpdater
+	)
+
+	if rolloutBlockUpdater, ok = obj.(RolloutBlockUpdater); !ok {
```
icanhazbroccoli: Consider skipping the 1st return argument assignment, i.e.:
```go
if _, ok := obj.(RolloutBlockUpdater); !ok {
    //...
}
```

pkg/controller/rolloutblock/rolloutblock_controller.go
```diff
@@ -0,0 +1,622 @@
...
+	if updaterType == NewRolloutBlockObject {
+		return c.syncNewRolloutBlockObject(key)
+	}
+
+	overridingKey := rolloutBlockUpdater.OverridingObjectKey
+	if len(key) == 0 || len(overridingKey) == 0 {
+		return nil
+	}
+
+	var err error = nil
```
icanhazbroccoli: No need for an explicit nil assignment:
```go
var err error
```
As it's going to be set to nil by default.

pkg/controller/rolloutblock/rolloutblock_controller.go
```diff
@@ -0,0 +1,622 @@
...
+	isDeletedObject := rolloutBlockUpdater.IsDeletedObject
+
+	if isDeletedObject {
```
icanhazbroccoli: Is it possible to get rid of this if-else statement and encode all the information into updaterType?

pkg/controller/rolloutblock/rolloutblock_controller.go
```diff
@@ -0,0 +1,622 @@
...
+	if err != nil {
+		return err
+	}
+
+	if ns == shipper.ShipperNamespace {
+		apps, err := c.applicationLister.List(labels.Everything())
+		if err != nil {
+			return err
+		}
+		if err = c.addMultipleApplicationsToRolloutBlocks(apps, key, rolloutBlock); err != nil {
```
icanhazbroccoli: Consider dropping word Multiple: Applications is already in a plural form. `apps` can be moved to the last position in the argument list and accepted as trailing argument: `...*shipper.Application`

pkg/controller/rolloutblock/rolloutblock_controller.go
```diff
@@ -0,0 +1,622 @@
...
+	if err != nil {
+		return err
+	}
+
+	rolloutBlock, err := c.rolloutBlockLister.RolloutBlocks(ns).Get(name)
+	if err != nil {
+		return err
+	}
+
+	if ns == shipper.ShipperNamespace {
```
icanhazbroccoli: You can preserve a corresponding lister function, like:
```go
var listerFunc <here comes it's type>
switch ns {
case shipper.ShipperNamespace:
    listerFunc = c.applicationLister.List
default:
    listerFunc := c.applicationLister.Applications(ns).List
}
This will allow you to drop this if-else statement completely
```

pkg/controller/rolloutblock/rolloutblock_controller.go
```diff
@@ -0,0 +1,622 @@
...
+		}
+		if err = c.addMultipleReleasesToRolloutBlocks(rels, key, rolloutBlock); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (c *Controller) addMultipleReleasesToRolloutBlocks(releases []*shipper.Release, rolloutBlockKey string, rolloutBlock *shipper.RolloutBlock) error {
```
icanhazbroccoli: Consider changing the method name to:
```go
addReleasesToRolloutBlocks(blockKey, block, ...*shipper.Release)
```
In this case you can reuse this method for adding a singular app to the block.

pkg/controller/rolloutblock/rolloutblock_controller.go
```diff
@@ -0,0 +1,622 @@
...
+	var relsStrings []string
```
icanhazbroccoli: I'd suggest calling it `relKeys`, or simply: `keys`

pkg/controller/rolloutblock/rolloutblock_controller.go
```diff
@@ -0,0 +1,622 @@
...
+	if err != nil {
+		return shippererrors.NewKubeclientUpdateError(rolloutBlock, err).
+			WithShipperKind("RolloutBlock")
+	}
+
+	return nil
+}
+
+func (c *Controller) addMultipleApplicationsToRolloutBlocks(applications []*shipper.Application, rolloutBlockKey string, rolloutBlock *shipper.RolloutBlock) error {
+	var appsStrings []string
```
icanhazbroccoli: Same as with releases: `appKeys` or `keys`

pkg/controller/rolloutblock/rolloutblock_controller.go
```diff
@@ -0,0 +1,622 @@
...
+
+		overrideRBs := strings.Split(overrideRB, ",")
+		for _, rbKey := range overrideRBs {
+			if rbKey == rolloutBlockKey {
+				appsStrings = append(appsStrings, appKey)
+			}
+		}
+	}
+
+	if len(appsStrings) == 0 {
```
icanhazbroccoli: This block is obsolete

pkg/controller/rolloutblock/rolloutblock_controller.go
```diff
@@ -0,0 +1,622 @@
...
+			continue
+		}
+
+		overrideRB, ok := app.GetAnnotations()[shipper.RolloutBlocksOverrideAnnotation]
+		if !ok {
+			continue
+		}
+
+		overrideRBs := strings.Split(overrideRB, ",")
+		for _, rbKey := range overrideRBs {
```
icanhazbroccoli: I believe this block needs a comment: what kind of override patterns we expect (don't hesitate to drop an example) and what's the expected result.

pkg/util/rolloutblock/rolloutblock.go
```diff
@@ -0,0 +1,58 @@
...
+
+	shipper "github.com/bookingcom/shipper/pkg/apis/shipper/v1alpha1"
+	shippererrors "github.com/bookingcom/shipper/pkg/errors"
+)
+
+// Return a boolean to decide if the application should override one of the
+// existing rollout blocks, a string representing the existing rollout blocks
+// that are not overridden, and an error if the application is trying to override
+// a non-existing rolloutblock
+func ShouldOverrideRolloutBlock(overrideRB string, nsRBs []*shipper.RolloutBlock, gbRBs []*shipper.RolloutBlock) (bool, string, error) {
```
icanhazbroccoli: It would be a bit more idiomatic (which means you're welcome to ignore this comment :-) ) to call this method simply `ShouldOverride`, so the entire invocation reads as: `rolloutblock.ShouldOverride` (note words rollout block are already in this sentence).

pkg/util/rolloutblock/rolloutblock.go
```diff
@@ -0,0 +1,58 @@
...
+	overrideRBs := strings.Split(overrideRB, ",")
+	if len(overrideRB) == 0 {
```
icanhazbroccoli: Obsolete block

pkg/util/rolloutblock/rolloutblock.go
```diff
@@ -0,0 +1,58 @@
...
+		overrideRBs = []string{}
+	}
+	RBs := append(nsRBs, gbRBs...)
```
icanhazbroccoli: consider calling it `rbs`, as a leading capital-char naturally implies on var public visibility

pkg/util/rolloutblock/rolloutblock.go
```diff
@@ -0,0 +1,58 @@
...
+
+	nonOverriddenRBs, err := Difference(RBs, overrideRBs)
+
+	return len(nonOverriddenRBs) == 0, strings.Join(nonOverriddenRBs, ", "), err
```
icanhazbroccoli: not sure if the second return argument is meant to be parsed somewhere, but it's concatenated with an extra space, which will make comma-splits all around to be broken. Better keep it consistent.

pkg/util/rolloutblock/rolloutblock.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+// Set difference: existingRBs - overrideRBs
+// Return all RolloutBlocks that are not overridden.
+// Return an InvalidRolloutBlockOverrideError if trying to override a
+// non-existing rolloutblock object
+func Difference(existingRBs []*shipper.RolloutBlock, overrideRBs []string) ([]string, error) {
```
icanhazbroccoli: Consider calling it `Diff`

pkg/util/rolloutblock/rolloutblock.go
```diff
@@ -0,0 +1,58 @@
...
+		existingRBdict[rbFullName] = true
+		if _, ok := overrideRBdict[rbFullName]; !ok {
+			diff = append(diff, rbFullName)
+		}
+	}
+
+	// Non-existing blocks enlisted in this annotation should not be allowed.
+	for _, item := range overrideRBs {
+		if _, ok := existingRBdict[item]; !ok {
+			glog.Infof("This rollout block %s does not exists", item)
```
icanhazbroccoli: 1. Consider adding verbosity here, like: glog.V(4).Infof(...)
2. Is there a chance it will be producing tons of non-actionable logs?

pkg/util/rolloutblock/rolloutblock.go
```diff
@@ -0,0 +1,58 @@
+package rolloutblock
```
icanhazbroccoli: Taking into account a massive split-concat pattern in your code I might suggest introducing an additional level of abstraction, like: RolloutBlockSet, which is effectively an ordered set of rollout blocks. It can encapsulate all diff, includes, excludes etc logic and simplify strings operations in the modules above.

pkg/util/string/string.go
```diff
@@ -0,0 +1,74 @@
+package string
```
icanhazbroccoli: As a followup to the comment above, this module might be gone as currently it mixes up distinct structures: some methods are string array specific, some expect arrays to be sets, and there is no implication on order. I'd suggest keeping this logic inside rollout block specific middleware and not touching strings at all.

pkg/webhook/webhook.go
```diff
@@ -175,3 +196,112 @@ func (c *Webhook) validateHandlerFunc(review *admission_v1beta1.AdmissionReview)
 		Allowed: true,
 	}
 }
+
+func (c *Webhook) validateRelease(request *admission_v1beta1.AdmissionRequest, release shipper.Release) error {
+	var err error
+	if request.Operation == kubeclient.Create {
+		err = c.shouldBlockRelease(release)
```
icanhazbroccoli: Not sure I understand what's happening here: validateRequest is supposed to ensure a release object validity and return an error if something's up. In this line the error is assigned to an error returned by shouldBlockRelease, which implies on some boolean-returning idempotent call. The method name doesn't really reflect on any validation happening inside though.

pkg/webhook/webhook.go
```diff
@@ -175,3 +196,112 @@ func (c *Webhook) validateHandlerFunc(review *admission_v1beta1.AdmissionReview)
...
+			err = c.validateOverrideRolloutBlockAnnotation(overrideRB, release.Namespace)
+		}
+	}
+	return err
+}
+
+func (c *Webhook) validateApplication(request *admission_v1beta1.AdmissionRequest, application shipper.Application) error {
+	var err error
+	if request.Operation == kubeclient.Create {
+		err = c.shouldBlockApplication(application)
```
icanhazbroccoli: Same as above: validation method returns should... method error.

pkg/webhook/webhook.go
```diff
@@ -175,3 +196,112 @@ func (c *Webhook) validateHandlerFunc(review *admission_v1beta1.AdmissionReview)
...
+	}
+
+	if overrideRolloutBlock {
+		return nil
+	}
+
+	return shippererrors.NewRolloutBlockError(eventMessage)
+}
+
+func (c *Webhook) existingRolloutBlocks(namespace string) ([]*shipper.RolloutBlock, []*shipper.RolloutBlock) {
```
icanhazbroccoli: Consider returning a new structure like RolloutBlocks with attributes like: Local and Global

pkg/webhook/webhook.go
```diff
@@ -175,3 +196,112 @@ func (c *Webhook) validateHandlerFunc(review *admission_v1beta1.AdmissionReview)
...
+	}
+	return nsRBs, gbRBs
+}
+
+func (c *Webhook) validateOverrideRolloutBlockAnnotation(overrideRB string, namespace string) error {
+	if len(overrideRB) == 0 {
+		return nil
+	}
+
+	re, err := regexp.Compile("^[a-zA-Z0-9/-]+/[a-zA-Z0-9/-]+$")
```
icanhazbroccoli: As it's a static regexp, consider using MustCompile

cmd/shipper-state-metrics/collector.go
```diff
@@ -322,6 +332,36 @@ func (ssm ShipperStateMetrics) collectClusters(ch chan<- prometheus.Metric) {
 	}
 }
 
+func (ssm ShipperStateMetrics) collectRolloutBlocks(ch chan<- prometheus.Metric) {
+	nss, err := getNamespaces(ssm.nssLister)
+	if err != nil {
+		glog.Warningf("collect Namespaces: %s", err)
```
icanhazbroccoli: I'd suggest using glog.Error* family of functions for errors: this is something that goes beyond fyi level of severity.

cmd/shipper-state-metrics/collector.go
```diff
@@ -322,6 +332,36 @@ func (ssm ShipperStateMetrics) collectClusters(ch chan<- prometheus.Metric) {
...
+		return
+	}
+
+	rolloutBlocks, err := ssm.rbLister.List(everything)
+	if err != nil {
+		glog.Warningf("collect RolloutBlocks: %s", err)
```
icanhazbroccoli: Same as above: Errorf

cmd/shipperctl/configurator/cluster.go
```diff
@@ -422,6 +422,7 @@ func (c *Cluster) FetchKubernetesCABundle() ([]byte, error) {
 }
 
 func (c *Cluster) CreateValidatingWebhookConfiguration(caBundle []byte, namespace string) error {
+	shipperValidatingWebhookServicePath := "/validate"
```
icanhazbroccoli: I'd suggest moving it to a constant

crd/examples/rolloutblock.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: shipper.booking.com/v1alpha1
```
icanhazbroccoli: Really good example. Good job!

pkg/apis/shipper/v1alpha1/types.go
```diff
@@ -8,7 +8,8 @@ import (
 )
 
 const (
-	ShipperNamespace = "shipper-system"
+	ShipperNamespace            = "shipper-system"
+	GlobalRolloutBlockNamespace = "global-rollout-blocks"
```
icanhazbroccoli: Minor suggestion: rename it to `rollout-block-global`. In this case if we introduce alternative namespaces to this one, the autocomplete-aware clients will still work and suggest all these namespaces.

pkg/controller/application/application_controller.go
```diff
@@ -24,6 +25,7 @@ import (
 	shippererrors "github.com/bookingcom/shipper/pkg/errors"
 	apputil "github.com/bookingcom/shipper/pkg/util/application"
 	releaseutil "github.com/bookingcom/shipper/pkg/util/release"
+	rolloutBlockOverride "github.com/bookingcom/shipper/pkg/util/rolloutblock"
```
icanhazbroccoli: I'd suggest keeping the name all-lower-case as it distinguishes them visually from variables.

pkg/controller/application/application_controller.go
```diff
@@ -238,6 +247,11 @@ func (c *Controller) syncApplication(key string) error {
 	if app.Annotations == nil {
 		app.Annotations = map[string]string{}
 	}
+	overrideRB, ok := app.Annotations[shipper.RolloutBlocksOverrideAnnotation]
+	if ok {
+		overrideRBs := rolloutBlockOverride.NewOverride(overrideRB)
```
icanhazbroccoli: Some minor concerns about naming: above we have a var called `overrideRB` and here is another one called overrideRBs. The current state of the code doesn't send an explicit signal about the difference between these 2 vars: i.e. what's their types, why another one is needed etc. One of the options I see here is to omit introduction of the new variable and assign the result of the stringification to the Annotations set right away, like:
```
app.Annotations[shipper.RolloutBlocksOverrideAnnotation] = rolloutBlockOverride.NewOverride(overrideRB).String()
```
Another option is to think of a more explicit type difference underlining. And, of course, the 3rd option is to ignore it :-)

pkg/controller/application/application_controller_test.go
```diff
@@ -542,6 +573,10 @@ func TestDeletingAbortedReleases(t *testing.T) {
 			Status:  corev1.ConditionFalse,
 			Message: fmt.Sprintf(ReleaseActiveMessageFormat, releaseBar.Name),
 		},
+		{
+			Type:	shipper.ApplicationConditionTypeRolloutBlock,
+			Status: corev1.ConditionFalse,
```
icanhazbroccoli: None of the test changes include enabled rollout block. I believe it might be a good idea to cover this case as well.

hbarkal: I separated the tests I wrote in a different pr: https://github.com/bookingcom/shipper/pull/119 
this is here just so the current tests will not fail :)

pkg/controller/application/rolloutblock.go
```diff
@@ -0,0 +1,52 @@
+package application
+
+import (
+	shipper "github.com/bookingcom/shipper/pkg/apis/shipper/v1alpha1"
+	apputil "github.com/bookingcom/shipper/pkg/util/application"
+	rolloutBlockOverride "github.com/bookingcom/shipper/pkg/util/rolloutblock"
+	corev1 "k8s.io/api/core/v1"
```
icanhazbroccoli: Conventionally, imports are split into newline-padded sections like: stdlib, externals, internals. I believe k8s one might be moved above into external dependencies section and be padded from shipper imports.

pkg/controller/application/rolloutblock.go
```diff
@@ -0,0 +1,52 @@
...
+
+func (c *Controller) processRolloutBlocks(app *shipper.Application, nsRBs, gbRBs []*shipper.RolloutBlock) bool {
+	overrideRB, ok := app.GetAnnotations()[shipper.RolloutBlocksOverrideAnnotation]
+	if !ok {
+		overrideRB = ""
+	}
+	appOverrideRBs := rolloutBlockOverride.NewOverride(overrideRB)
+	rbs := append(nsRBs, gbRBs...)
+	existingRBs := rolloutBlockOverride.NewOverrideFromRolloutBlocks(rbs)
+	nonExistingRbs := appOverrideRBs.Diff(existingRBs)
```
icanhazbroccoli: A comment on naming: a nonExistingRb makes me think it's up for creation, but from reading the code below it seems it's up for cleanup. My suggestion is to indicate that these items are outdated and are up for removal by renaming this var to something like: `obsoleteRbs` (or something around this sentiment).

pkg/controller/release/rolloutblock.go
```diff
@@ -0,0 +1,55 @@
+package release
```
icanhazbroccoli: The file name might indicate it's an extension to scheduler

pkg/controller/release/rolloutblock.go
```diff
@@ -0,0 +1,55 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/runtime"
+
+	shipper "github.com/bookingcom/shipper/pkg/apis/shipper/v1alpha1"
+	rolloutBlockOverride "github.com/bookingcom/shipper/pkg/util/rolloutblock"
+)
+
+func (s *Scheduler) processRolloutBlocks(rel *shipper.Release) (bool, string) {
+	relOverrideRB, ok := rel.Annotations[shipper.RolloutBlocksOverrideAnnotation]
```
icanhazbroccoli: This code looks quite identical to https://github.com/bookingcom/shipper/pull/106/files#diff-3b18da970d68d60671e98b680f1b9d06R10
Do you think there is still some space for DRY'ing?

pkg/controller/rolloutblock/rolloutblock_controller.go
```diff
@@ -0,0 +1,285 @@
...
+
+	rolloutBlockInformer.Informer().AddEventHandler(
+		cache.ResourceEventHandlerFuncs{
+			AddFunc: controller.onAddRolloutBlock,
+		})
+
+	return controller
+}
+
+func (c *Controller) enqueueRelease(obj interface{}) {
```
icanhazbroccoli: Despite this function says enqueueRelease, I'm not sure it does this explicitly

pkg/controller/rolloutblock/rolloutblock_controller.go
```diff
@@ -0,0 +1,285 @@
...
+	rel, ok := obj.(*shipper.Release)
+	if !ok {
+		runtime.HandleError(fmt.Errorf("not a shipper.Release: %#v", obj))
+		return
+	}
+
+	overrideRB, _ := rel.GetAnnotations()[shipper.RolloutBlocksOverrideAnnotation]
+	overrideRBs := rolloutBlockOverride.NewOverride(overrideRB)
+	for rbKey := range overrideRBs {
```
icanhazbroccoli: I wonder if there is a specific reason for plural enqueueing, not singular stringified?

pkg/controller/application/application_controller_test.go
icanhazbroccoli: Oh, that totally makes sense, thanks! Sorry for some extra noise!

juliogreff:
I see quite some unrelated changes here due to master diverging over time. @hbarkal can you please rebase so we're sure everything still works?

Issue #77, opened 2019-03-20
Shipper Doesn't Work With Resyncs Disabled


parhamdoustdar:
If you run Shipper with resyncs disabled, there are cases where Shipper will get stuck while taking a release forward.

This is because sometimes we rely on the resync to happen, and we don't do as much as we can in each sync loop.

We haven't dug into this yet, so the best way to reproduce this is to run the end-to-end tests while running Shipper with the resync period of 0 to disable all resyncs, and go from there.

juliogreff:
I followed the suggestion of just running the end-to-end tests. Much to my surprise, they all passed. Unfortunately, that seems to be because in current master there's no way to completely disable resyncs. I fixed that in 7ae1c636b93eeb680faea8fedfa69e228dd43ad6.

With those changes, I get one failing test:

```
--- FAIL: TestRolloutAbort (73.78s)
    e2e_test.go:751: waiting for release "my-test-app-77cf3473-0" took 1.597339649s
    e2e_test.go:840: waiting for completion of "my-test-app-77cf3473-0" took 18.399953573s
    e2e_test.go:617: waiting for contender release to appear after editing app "my-test-app"
    e2e_test.go:751: waiting for release "my-test-app-3037a9fd-0" took 1.598389769s
    e2e_test.go:624: setting contender release "my-test-app-3037a9fd-0" targetStep to 0
    e2e_test.go:627: waiting for contender release "my-test-app-3037a9fd-0" to achieve waitingForCommand for targetStep 0
    e2e_test.go:761: release strategy state transition: "" -> "release \"my-test-app-3037a9fd-0\" has no achievedStep reported yet"
    e2e_test.go:761: release strategy state transition: "release \"my-test-app-3037a9fd-0\" has no achievedStep reported yet" -> "{installation: False, capacity: False, traffic: False, command: True}"
    e2e_test.go:819: waiting for command took 10.398270118s
    e2e_test.go:633: checking that incumbent "my-test-app-77cf3473-0" has 4 pods and contender "my-test-app-3037a9fd-0" has 1 pods (strategy step 0 -- 100/1)
    e2e_test.go:624: setting contender release "my-test-app-3037a9fd-0" targetStep to 1
    e2e_test.go:627: waiting for contender release "my-test-app-3037a9fd-0" to achieve waitingForCommand for targetStep 1
    e2e_test.go:761: release strategy state transition: "" -> "{installation: False, capacity: True, traffic: False, command: False}"
[...]

PR #117, opened 2019-07-11
Handle errors from cleanUpReleasesForApplication

juliogreff:
This addresses another part of #77, trying to make shipper work with
resyncs disabled.

Before this change, cleaning up the list of releases was "fire and
forget": we wouldn't check if any errors happened, and would just keep
on trucking as if nothing happened. This is fine when you can expect
resyncs to pick up the slack, but since we're specifically trying to
avoid that, we need to check whenever errors happen, and retry
appropriately.

### DIFF
--- a/pkg/controller/application/application_controller.go
+++ b/pkg/controller/application/application_controller.go
@@ -358,12 +358,6 @@ func (c *Controller) processApplication(app *shipper.Application) error {
 	// Required by subsequent calls to GetContender and GetIncumbent.
 	appReleases = releaseutil.SortByGenerationDescending(appReleases)
 
-	// clean up excessive releases regardless of exit path
-	defer func() {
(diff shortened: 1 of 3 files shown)


pkg/controller/application/application_controller.go
```diff
@@ -459,18 +460,24 @@ func (c *Controller) processApplication(app *shipper.Application) error {
 	}
 
 	apputil.SetHighestObservedGeneration(app, highestObserved)
+
+	app.Status.History = apputil.ReleasesToApplicationHistory(appReleases)
+	if err = c.cleanUpReleasesForApplication(app, appReleases); err != nil {
+		return err
+	}
+
```
icanhazbroccoli: These 3 added blocks are fairly similar to each other. I wonder if it feels like apputil might be the right place for a new method incorporating this logic?

PR #119, opened 2019-07-15
Hbarkal/1 rollout blocks tests

hbarkal:
Tests for [RolloutBlock](https://github.com/bookingcom/shipper/pull/106/) 
Testing blocking functionality, overrides, updating RolloutBlock status


pkg/controller/release/release_controller_test.go
```diff
@@ -292,6 +314,11 @@ func (f *fixture) buildIncumbent(namespace string, relName string, replicaCount
 		f.t.Fatalf("The fixture is missing at least 1 Cluster object")
 	}
 
+	rolloutblocksOverrides, ok := app.Annotations[shipper.RolloutBlocksOverrideAnnotation]
```
icanhazbroccoli: Annotations is defined like: `map[string]string`. In this case if `v, ok := map[k]` is returned with no-key flag (ok === false), v will be defined and have a zero-value (for string type it's ""). From that perspective, these 4 lines can be stated as:
```
rolloutBlockOverrides := app.Annotations[shipper.RolloutBlockOverrideAnnotation]
```
This pattern is very common in https://github.com/bookingcom/shipper/pull/106 as well, but I left no comments on it.

PR #127, opened 2019-07-24
Rate limiting the release objects that on update are unchanged

hbarkal:
Addressing the spike on resyncs, rate limiting the release objects that on update are unchanged.

We want to space out release objects over the time between resyncs. We want to do that only for release objects that are unchanged, in order to not introduce latency to updating release object that require an update.

### DIFF
--- a/pkg/controller/release/release_controller.go
+++ b/pkg/controller/release/release_controller.go
@@ -135,6 +135,21 @@ func NewController(
 		cache.ResourceEventHandlerFuncs{
 			AddFunc: controller.enqueueRelease,
 			UpdateFunc: func(oldObj, newObj interface{}) {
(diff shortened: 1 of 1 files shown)


pkg/controller/release/release_controller.go
```diff
@@ -135,6 +135,13 @@ func NewController(
 		cache.ResourceEventHandlerFuncs{
 			AddFunc: controller.enqueueRelease,
 			UpdateFunc: func(oldObj, newObj interface{}) {
+				oldRel, oldOk := oldObj.(*shipper.Release)
```
icanhazbroccoli: I suggest to let it bail out right away of oldOk is false. Same for newOk. Otherwise the if-statement looks overoptimistic.

PR #135, opened 2019-07-26
Give more information for errors of failed chart repo index

hbarkal:
Add an error for no chart repo to give more information to the user
This error will be used when we failed to fetch chart repo, and there's no previous index in cache for fallback

### DIFF
--- a/pkg/chart/repo/repo.go
+++ b/pkg/chart/repo/repo.go
@@ -81,6 +81,19 @@ func (r *Repo) Start(stopCh <-chan struct{}) {
 func (r *Repo) refreshIndex() error {
 	data, err := r.fetcher(r.indexURL)
 	if err != nil {
+		_, cacheErr := r.cache.Fetch("index.yaml")
+		if cacheErr != nil {
+			multiError := shippererrors.NewMultiError()
+			multiError.Append(
+				shippererrors.NewChartRepoIndexError(
+					fmt.Errorf("failed to fetch %q: %v", r.indexURL, err),
+				))
+			multiError.Append(
+				shippererrors.NewNoCachedChartRepoIndexError(
+					fmt.Errorf("failed to fetch %q: %v", r.indexURL, cacheErr),
+				))
+			return multiError
+		}
 		return shippererrors.NewChartRepoIndexError(
 			fmt.Errorf("failed to fetch %q: %v", r.indexURL, err),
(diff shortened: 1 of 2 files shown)


pkg/chart/repo/handlers.go
```diff
@@ -22,6 +22,10 @@ func ResolveChartVersionFunc(c *Catalog) ChartVersionResolver {
 
 		if _, err := repo.RefreshIndex(); err != nil {
 			glog.Warningf("failed to refresh repo[%s] index: %s", chartspec.RepoURL, err)
+			switch err.(type) {
+			case errors.NoCachedChartRepoIndexError:
+				return nil, err
```
icanhazbroccoli: I suggest adding `default:` statement (even an empty one if a fall through is meant here) so the alternative action is explicitly visible.

pkg/chart/repo/repo.go
```diff
@@ -61,6 +61,13 @@ func (r *Repo) RefreshIndex() (*repo.IndexFile, error) {
 
 	data, err := r.fetcher(indexURL)
 	if err != nil {
+		_, cacheErr := r.cache.Fetch("index.yaml")
+		if cacheErr != nil {
```
icanhazbroccoli: Do we care about cacheErr contents? It seems it's lost after a non-nil check

hbarkal: The error that we are printing to logs is the response code from the server when trying to fetch the chart (for example: "unsuccessful response code: 500 Internal Server Error (500)"). cacheErr will hold the error for trying to fetch from cache (for example: "open /tmp/chart-cache/https-charts.prod.booking.com-charts/index.yaml: no such file or directory")
I can add cacheErr if you think it should be there as well

juliogreff:
It looks like we have conflicts from when #143 got merged. @hbarkal can you rebase?

juliogreff:
@hbarkal we have conflicts again after recent refactorings on repo index fetching. Can you rebase again?

pkg/chart/repo/repo.go
icanhazbroccoli: I see your point. It seems to be a composite error: we got err and cacheErr, both worth reporting. I might suggest using shippererrors.MultiError in that case so we don't miss the actual err reason contained in err.

Issue #62, opened 2018-12-21
chart version range specifications


kanatohodets:
In order to keep a Chart ecosystem healthy, it is important for some range of Chart updates to be applied quickly and consistently. Encoding a specific version into Application objects and then updating it with every change means that new versions are unlikely to be applied across the whole population of Application objects: the Chart maintainers might release a new version, but only some Application owners will action the upgrade, which represents toil for them.

To address this, we think it makes sense for Shipper to develop the ability to work with semver version range specifications for Charts. In this way, Application owners can specify a version constraint that they're comfortable with (for example, upwards float on minor or bugfix versions), and get those updates automatically.

### Approach

Allow a version range specification as part of the `version` key of the `chart` field in Application objects; for example, a [floating patch specification](https://github.com/Masterminds/semver#tilde-range-comparisons-patch): `~1.1.0`. When a Release is created as part of a rollout, Shipper should resolve that range specification to a concrete, pinned version, such as `1.1.6`. The Release should get that pinned version in the `version` key. 

Release will not support version range constraints: Releases must always have a concrete version. This ensures that aborts and rollbacks always go back to a predictable version. Version range resolution is a feature of the higher level Application interface.

#### Become a real Helm repo client
In order to resolve the version range spec, Shipper will need to become a fully fledged Helm repo client which processes `index.yaml`. Right now it just combines the `version` and `repoUrl` and appends `.tgz` to try to download a Chart directly, which is incorrect behavior. If we were to keep this behavior, `repoUrl` should be renamed to `chartUrl`.

#### Speed
Shipper should check for an updated `index.yaml` as soon as it needs to do a rollout. This allows CI/CD pipelines to push a new chart and then immediately deploy it.

#### Robustness to repo unavailability
Shipper currently caches Charts aggressively. This is in order to prevent rollouts with a previously-deployed Chart from failing due to a broken Chart repository. We want to maintain this property as much as possible: we'd like to avoid going over the wire to contact the chart repo unless absolutely required. As such:

1. We always try to use a cached Chart tarball when dealing with concrete versions.
2. If the `index.yaml` fetch fails when resolving a version range specification, we should try to fall back to a cached `index.yaml`.

This means that a version range resolution _may_ not return the latest possible version if the Chart repo is down. In this case Shipper should indicate the resolution failure or fallback using a Condition on the Application (or similar).

isutton:
[...]

PR #143, opened 2019-08-02
Reworked chart version range resolution strategy refs #62

icanhazbroccoli:
This change is aiming to address the problem we encountered during
exploiting release-0.5. In this release we introduced helm chart range
functionality, where an application.spec.template.chart.version could
contain a SemVer constraint and be resolved to a material chart version.
In the original implementation we decided to keep
application.spec.template.chart.version as given and preserve resolved
chart version in release.spec.environment.chart.version. Under these
circumstances we broke 2 things: normal rollout and rollbacks.

Rollouts were broken due to an always-mismatch between env hashes of the
application and it's contender: due to the difference between the
versions, env hashes were never the same and therefore Shipper was
continuously re-creating new releases. On top of that, an attempt to
re-rollout the same app with a constraint would fail as well even if we
had fixed the hash problem: they would be the same, which Shipper
recofnises as a no-op.

Rollbacks were breaking the contract due to the environment back-copy
operations: during a rollback Shipper copies contender spec back to the
application. This meant we're loosing the initial information about the
original SemVer constraint.

This change takes a different approach. In essence,
application.spec.template.chart.version is being updated on the release
stage: the chart version is being materialised and preserved in the
application spec instead of the initial constraint. The constraint is
now preserved in application annotations among with the chart name and
the resolved version. These 3 components are being used to detect
whether a specific chart version has already been resolved and whether
it's consistent with the one specified in the spec. In this case the
rest of the algorithm stays in place: releases are being created with
the updated materialised chart spec, and rollbacks don't loose any data.


pkg/controller/application/application_controller_utils.go
```diff
@@ -115,6 +108,11 @@ func identicalEnvironments(envs ...shipper.ReleaseEnvironment) bool {
 		return true
 	}
 
+	// chartVersions := make([]string, 0, len(envs))
+	// for _, env := range envs {
```
juliogreff: Commented out code?

icanhazbroccoli: <3 Totally forgot about it, thanks!

PR #148, opened 2019-08-08
Define better printing Columns for Shipper CRDs

### LINKED ISSUE #79: Define Better Printing Columns For Our Custom Resources
Shipper CRDs are currently not taking advantage of the capabilities added in Kubernetes 1.13 for defining printing columns. These are great for improving the user experience of Shipper, so it's important for us to define and surface useful information at each level.

#67 already discusses this for Release objects, but we need to define this for other CRDs such as Application and target objects as well.

juliogreff:
Closes #79.

Now, when using `kubectl`, one can expect the following outputs:

```
$ kubectl get app
NAME        LATEST RELEASE         ROLLING OUT   AGE
snowflake   snowflake-306ba2a4-0   False         22h

$ kubectl get rel
NAME                   STEP      ERRORS            AGE
snowflake-1578afe1-0   vanguard  BrokenChartSpec   10m

$ kubectl get rel -o wide
NAME                   STEP      ERRORS   AGE    CLUSTERS
snowflake-1779e269-0   full on            18m    local
```

### DIFF
--- a/crd/Application-crd.yaml
+++ b/crd/Application-crd.yaml
@@ -4,6 +4,19 @@ metadata:
   # name must match the spec fields below, and be in the form: <plural>.<group>
   name: applications.shipper.booking.com
 spec:
+  additionalPrinterColumns:
+  - JSONPath: .status.history[-1]
+    description: The application's latest release.
+    name: Latest release
+    type: string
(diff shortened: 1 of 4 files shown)


juliogreff:
Release controller conditions are a bit in flux right now, so I don't think that this PR will land as it is, so I'll close this and reopen when we're ready to move forward.

PR #151, opened 2019-08-08
Rollout Blocks

### LINKED ISSUE #1: Rollout blocks

hbarkal:
This introduces RolloutBlocks as a CRD in shipper.
Blocking rollouts locally or globally is possible using a RolloutBlock object in the relevant namespace.

Closes #1


docs/user/rolling-out.rst
```diff
@@ -174,3 +174,155 @@ And confirm that Shipper believes this rollout to be done:
...
+    metadata:
+      name: ui
+      namespace: frontend
+    spec:
+      # ... spec omitted
+    status:
+      conditions:
+      - type: RolloutBlock
+        status: True
+        reason: rollout-blocks-global/dns-outage
```
juliogreff: [K8s conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status) say this about `reason`:

>In condition types, and everywhere else they appear in the API, Reason is intended to be a one-word, CamelCase representation of the category of cause of the current status


On the other hand, saying that a `RolloutBlock` reason is... `RolloutBlock`... it's a bit weird, so I wonder if this is the correct name for this condition.

docs/user/rolling-out.rst
```diff
@@ -174,3 +174,155 @@ And confirm that Shipper believes this rollout to be done:
...
+
+``output wide``
+---------------
+
+This will show all information about all rollout blocks in the namsespace (default if not specify,
+`rollout-blocks-global` for all global RolloutBlocks ,`--all-namespaces` for all rollout blocks)
+
+.. code-block:: shell
+
+    $ kubectl -n rollout-blocks-global get rb -o wide
```
juliogreff: Really nice usage of `wide` here <3

pkg/controller/application/application_controller.go
```diff
@@ -238,6 +247,10 @@ func (c *Controller) syncApplication(key string) error {
 	if app.Annotations == nil {
 		app.Annotations = map[string]string{}
 	}
+	overrideRB, ok := app.Annotations[shipper.RolloutBlocksOverrideAnnotation]
+	if ok {
+		app.Annotations[shipper.RolloutBlocksOverrideAnnotation] = rolloutblock.NewOverride(overrideRB).String()
```
juliogreff: This looks like it's taking a string, converting to some object, just to have it back as a string of the same form (or at least most of the times, I realize it removes empty overrides). Am I missing other uses for this?

pkg/controller/application/rolloutblock.go
```diff
@@ -0,0 +1,53 @@
...
+		c.recorder.Event(app, corev1.EventTypeWarning, "Non Existing RolloutBlock", obsoleteRbs.String())
+	}
+
+	nonOverriddenRBs := existingRBs.Diff(appOverrideRBs)
+	shouldBlockRollout := len(nonOverriddenRBs) != 0
+
+	if shouldBlockRollout {
+		c.recorder.Event(app, corev1.EventTypeWarning, "RolloutBlock", nonOverriddenRBs.String())
+	} else if len(appOverrideRBs) > 0 {
+		c.recorder.Event(app, corev1.EventTypeNormal, "Overriding RolloutBlock", appOverrideRBs.String())
```
juliogreff: Same issue with `reason` here in `Event` (the 3rd parameter)

pkg/controller/release/release_controller.go
```diff
@@ -435,6 +459,22 @@ func (c *Controller) enqueueRelease(obj interface{}) {
 	c.releaseWorkqueue.Add(key)
 }
 
+func (c *Controller) enqueueReleaseRateLimited(obj interface{}) {
```
juliogreff: Not that I mind (as I have already approved it anyway), but these changes are already present in a different PR: #127

pkg/controller/release/release_controller.go
```diff
@@ -517,6 +557,10 @@ func reasonForReleaseCondition(err error) string {
 		return "BrokenChartSpec"
 	case shippererrors.WrongChartDeploymentsError:
 		return "WrongChartDeployments"
+	case shippererrors.InvalidRolloutBlockOverrideError:
+		return "InvalidRolloutBlockOverride"
```
juliogreff: Is this an error that can actually happen during a release sync?

pkg/controller/release/scheduler_rolloutblock.go
```diff
@@ -0,0 +1,52 @@
...
+		runtime.HandleError(fmt.Errorf("failed to list rollout block objects: %s", err))
+	}
+
+	gbRBs, err := s.rolloutBlockLister.RolloutBlocks(shipper.GlobalRolloutBlockNamespace).List(labels.Everything())
+	if err != nil {
+		runtime.HandleError(fmt.Errorf("failed to list rollout block objects: %s", err))
+	}
+
+	existingRBs := rolloutblock.NewOverrideFromRolloutBlocks(append(nsRBs, gbRBs...))
+	obsoleteRbs := relOverrideRBs.Diff(existingRBs)
```
juliogreff: This bit looks very similar to what happens in the application controller. Maybe we should DRY it up?

pkg/errors/rolloutblock.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"fmt"
+)
+
+type InvalidRolloutBlockOverrideError struct {
+	RolloutBlockName string
+}
+
+func (e InvalidRolloutBlockOverrideError) Error() string {
+	return fmt.Sprintf("rollout block with name %s does not exists",
```
juliogreff: "does not exist" ;)

pkg/webhook/webhook.go
```diff
@@ -175,3 +192,102 @@ func (c *Webhook) validateHandlerFunc(review *admission_v1beta1.AdmissionReview)
...
+	err = c.validateOverrideRolloutBlockAnnotation(overrideRBs, release.Namespace)
+	if err != nil {
+		return err
+	}
+
+	if request.Operation == kubeclient.Create {
+		err = c.processRolloutBlocks(release.Namespace, overrideRBs)
+	}
+
+	if request.Operation == kubeclient.Update {
```
juliogreff: These two if statements look mutually exclusive. I'd prefer either a switch, or an if-else-if, for the sake of explicitness of intent.

pkg/controller/application/application_controller.go
hbarkal: I think mainly to have it in the same order every time

pkg/controller/release/release_controller.go
hbarkal: Yes, some tests fail without it (as they cause updating a release which triggers the gazillion retries)

pkg/controller/release/scheduler_rolloutblock.go
hbarkal: Tried, it did not go well. It is *very* similar, but removing the duplications removed only 4 lines and introduce ambiguity...

docs/user/rolling-out.rst
```diff
@@ -174,3 +174,155 @@ And confirm that Shipper believes this rollout to be done:
...
+.. code-block:: yaml
+    apiVersion: shipper.booking.com/v1alpha1
+    kind: RolloutBlock
+    metadata:
+      name: dns-outage
+      namespace: rollout-blocks-global # for global rollout block. for a local one use the correct namespace.
+    spec:
+      message: DNS issues, troubleshooting in progress
+      author:
+        type: user
```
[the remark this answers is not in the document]
hbarkal: It was suggested in the RFC, and I think it opens an opportunity for an automated rollout block, maybe...

docs/user/rolling-out.rst
```diff
@@ -174,3 +174,155 @@ And confirm that Shipper believes this rollout to be done:
...
+        name: jdoe # This indicates that a rollout block was put in place by user 'jdoe'
+
+Copy this to a file called ``globalRolloutBlock.yaml`` and apply it to our Kubernetes cluster:
+
+.. code-block:: shell
+
+    $ kubectl apply -f globalRolloutBlock.yaml
+
+This will create a Global *RolloutBlock* object.
+In order to create a namespace rollout block, simply state the relevant namespace in the yaml file.
```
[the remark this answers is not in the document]
hbarkal: Global rolloutblock will live in `rollout-blocks-global`, namespaced rolloutblocks will live in the same namespace they block

pkg/apis/shipper/v1alpha1/types.go
```diff
@@ -101,7 +101,7 @@ const (
 	ApplicationConditionTypeReleaseSynced ApplicationConditionType = "ReleaseSynced"
 	ApplicationConditionTypeAborting      ApplicationConditionType = "Aborting"
 	ApplicationConditionTypeRollingOut    ApplicationConditionType = "RollingOut"
-	ApplicationConditionTypeRolloutBlock  ApplicationConditionType = "RolloutBlock"
+	ApplicationConditionTypeRolloutBlock  ApplicationConditionType = "Blocked"
```
juliogreff: Maybe we should now rename this constant to `ApplicationConditionTypeBlocked`?

cmd/shipper-state-metrics/collector.go
```diff
@@ -322,6 +332,36 @@ func (ssm ShipperStateMetrics) collectClusters(ch chan<- prometheus.Metric) {
...
+		rbsPerNamespace[rolloutBlock.Namespace]++
+	}
+
+	glog.V(4).Infof("RolloutBlocks: %v", rbsPerNamespace)
+
+	for _, ns := range nss {
+		n, ok := rbsPerNamespace[ns.Name]
+		if !ok {
+			n = 0
+		}
```
icanhazbroccoli: Minor: not needed: you can rely on golang map zero-value return, in this case it's going to be 0.0

cmd/shipper-state-metrics/collector.go
```diff
@@ -322,6 +332,36 @@ func (ssm ShipperStateMetrics) collectClusters(ch chan<- prometheus.Metric) {
...
+
+	rolloutBlocks, err := ssm.rbLister.List(everything)
+	if err != nil {
+		glog.Errorf("collect RolloutBlocks: %s", err)
+		return
+	}
+
+	rbsPerNamespace := make(map[string]float64)
+	for _, rolloutBlock := range rolloutBlocks {
+		rbsPerNamespace[rolloutBlock.Namespace]++
```
icanhazbroccoli: Minor: if I understand correctly, float64 is chosen because prometheus gauge wants it. In the other hand, the semantics of this value is a counter, which means an integer-like value is more natural for operations like ++. See https://stackoverflow.com/questions/8561393/is-using-increment-operator-on-floats-bad-style for more details. As I said, this is a matter of polishing the semantics of this method, so, not a blocker to me.

cmd/shipper-state-metrics/collector.go
```diff
@@ -337,7 +377,7 @@ func getNamespaces(lister kubelisters.NamespaceLister) ([]*corev1.Namespace, err
 		return nil, err
 	}
 
-	nsBlacklist := []string{"kube-system", "kube-public", "kube-dns", shipper.ShipperNamespace}
+	nsBlacklist := []string{"kube-system", "kube-public", "kube-dns", shipper.ShipperNamespace, shipper.GlobalRolloutBlockNamespace}
```
icanhazbroccoli: I'd' suggest leaving a comment on the meaning of this group of namespaces: what do they define and what are the consequences if a namespace is mentioned here

cmd/shipper/main.go
```diff
@@ -510,13 +513,39 @@ func startTrafficController(cfg *cfg) (bool, error) {
...
 		return false, nil
 	}
 
-	c := webhook.NewWebhook(cfg.webhookBindAddr, cfg.webhookBindPort, cfg.webhookKeyPath, cfg.webhookCertPath)
+	c := webhook.NewWebhook(
+		cfg.webhookBindAddr,
+		cfg.webhookBindPort,
+		cfg.webhookKeyPath,
+		cfg.webhookCertPath,
+		buildShipperClient(cfg.restCfg, rolloutblock.AgentName, cfg.restTimeout))
```
icanhazbroccoli: Let's move `buildShipperClient` outside of this call. Looking at the argument list (say, the timeout value), it makes me think the building can fail eventually and we might consider checking err-nilness before creating a webhook

pkg/apis/shipper/v1alpha1/types.go
```diff
@@ -43,6 +44,8 @@ const (
 	SecretChecksumAnnotation             = "shipper.booking.com/cluster-secret.checksum"
 	SecretClusterSkipTlsVerifyAnnotation = "shipper.booking.com/cluster-secret.insecure-tls-skip-verify"
 
+	RolloutBlocksOverrideAnnotation = "shipper.booking.com/block.override"
```
icanhazbroccoli: Proposed change: shipper.booking.com/rollout-block.override

pkg/client/clientset/versioned/typed/shipper/v1alpha1/fake/fake_rolloutblock.go
```diff
@@ -0,0 +1,124 @@
...
+
+	if obj == nil {
+		return nil, err
+	}
+	return obj.(*v1alpha1.RolloutBlock), err
+}
+
+// Update takes the representation of a rolloutBlock and updates it. Returns the server's representation of the rolloutBlock, and an error, if there is any.
+func (c *FakeRolloutBlocks) Update(rolloutBlock *v1alpha1.RolloutBlock) (result *v1alpha1.RolloutBlock, err error) {
+	obj, err := c.Fake.
```
icanhazbroccoli: Not sure about this line break. Let's' keep method invocations on the same line.

pkg/client/clientset/versioned/typed/shipper/v1alpha1/fake/fake_rolloutblock.go
```diff
@@ -0,0 +1,124 @@
...
+	if obj == nil {
+		return nil, err
+	}
+	return obj.(*v1alpha1.RolloutBlock), err
+}
+
+// UpdateStatus was generated because the type contains a Status member.
+// Add a +genclient:noStatus comment above the type to avoid generating UpdateStatus().
+func (c *FakeRolloutBlocks) UpdateStatus(rolloutBlock *v1alpha1.RolloutBlock) (*v1alpha1.RolloutBlock, error) {
+	obj, err := c.Fake.
```
icanhazbroccoli: Same as above: line break

pkg/controller/application/application_controller.go
icanhazbroccoli: It's a stable annotation stringer, yes. It does de-duplication and sorting inside. The code was full of strings.Split() and strings.Join() + de-duplication code all around. The idea is to move it to an object.

juliogreff: Sounds like a comment pointing that out would be in order, then, as it's non obvious and relatively important

cmd/shipper/main.go
hbarkal: And incase `buildShipperClient` returns a nil object, should we raise an error?

juliogreff: `buildShipperClient` (and its friend buildKubeClient) actually panic on error. Unless we plan on changing that (and I think that should probably go on a different PR), the call here is appropriate.

pkg/controller/application/rolloutblock.go
```diff
@@ -0,0 +1,54 @@
...
+func (c *Controller) processRolloutBlocks(app *shipper.Application, nsRBs, gbRBs []*shipper.RolloutBlock) bool {
+	appOverrideRBs := rolloutblock.NewOverride(app.GetAnnotations()[shipper.RolloutBlocksOverrideAnnotation])
+	rbs := append(nsRBs, gbRBs...)
+
+	existingRBs := rolloutblock.NewOverrideFromRolloutBlocks(rbs)
+	obsoleteRbs := appOverrideRBs.Diff(existingRBs)
+
+	if len(obsoleteRbs) > 0 {
+		for o := range obsoleteRbs {
+			c.removeRolloutBlockFromAnnotations(appOverrideRBs, o, app)
```
icanhazbroccoli: Proposed todo: move it to rolloutblockutils

pkg/controller/release/release_controller.go
```diff
@@ -135,6 +142,21 @@ func NewController(
 		cache.ResourceEventHandlerFuncs{
 			AddFunc: controller.enqueueRelease,
 			UpdateFunc: func(oldObj, newObj interface{}) {
+				oldRel, ok := oldObj.(*shipper.Release)
+				if !ok {
+					return
```
icanhazbroccoli: Shall we do runtime.HandleError here?

pkg/controller/application/rolloutblock.go
hbarkal: Can it be done in rolloutblockutil? It is updating the annotation of an application object

pkg/webhook/webhook.go
```diff
@@ -175,3 +192,102 @@ func (c *Webhook) validateHandlerFunc(review *admission_v1beta1.AdmissionReview)
...
+
+func (c *Webhook) validateApplication(request *admission_v1beta1.AdmissionRequest, application shipper.Application) error {
+	var err error
+	overrideRBs := rolloutblock.NewOverride(application.Annotations[shipper.RolloutBlocksOverrideAnnotation])
+	err = c.validateOverrideRolloutBlockAnnotation(overrideRBs, application.Namespace)
+	if err != nil {
+		return err
+	}
+
+	if request.Operation == kubeclient.Create {
```
juliogreff: I believe we should also validate against rollout blocks if we're trying to roll out an existing application. Currently, changing any of the values in the spec triggers rollout blocks, but the webhook does not return an error.

pkg/controller/rolloutblock/rolloutblock_controller.go
```diff
@@ -0,0 +1,313 @@
...
+			AddFunc: controller.enqueueApplicationBlock,
+			UpdateFunc: func(oldObj, newObj interface{}) {
+				controller.enqueueApplicationBlock(newObj)
+			},
+			DeleteFunc: controller.enqueueApplicationBlock,
+		})
+
+	rolloutBlockInformer.Informer().AddEventHandler(
+		cache.ResourceEventHandlerFuncs{
+			AddFunc: controller.onAddRolloutBlock,
```
juliogreff: I wonder if we shouldn't also listen to delete events here, and remove the rollout block being removed from all overrides in the apps and releases.

Currently, if a rollout block that was overriden gets deleted, and one tries to edit the app or release that overrode it, it'll error out with "rollout block with name foo/bar does not exist"

pkg/webhook/webhook.go
```diff
@@ -175,3 +192,102 @@ func (c *Webhook) validateHandlerFunc(review *admission_v1beta1.AdmissionReview)
...
+		for _, item := range nsRBList.Items {
+			nsRBs = append(nsRBs, &item)
+		}
+	}
+	if gbRBList, err := c.shipperClientset.ShipperV1alpha1().RolloutBlocks(shipper.GlobalRolloutBlockNamespace).List(meta_v1.ListOptions{}); err == nil {
+		for _, item := range gbRBList.Items {
+			gbRBs = append(gbRBs, &item)
+		}
+	}
+	return append(nsRBs, gbRBs...)
```
juliogreff: Instead of building two separate slices, and then appending them together, I guess you could just append all rollout blocks to the same, shared slice.

pkg/webhook/webhook.go
hbarkal: So the Update case is a bit tricky, and even though webhook doesn't validate againts rollout block in updates, application and release controller will still maintain blocking mechanism. Webhook should accept the change in case the change it in annotation (like removing an override in order to block the object). If we validate against rollout blocks in case on an update as well, webhook will reject cases like these. I can add a check to see if it's a case of removed annotation, it might cause the validating function to be a bit weird looking but I think it will be just fine...

pkg/webhook/webhook.go
```diff
@@ -175,3 +192,102 @@ func (c *Webhook) validateHandlerFunc(review *admission_v1beta1.AdmissionReview)
...
+
+	if request.Operation == kubeclient.Create {
+		err = c.processRolloutBlocks(application.Namespace, overrideRBs)
+	}
+
+	return err
+}
+
+func (c *Webhook) processRolloutBlocks(namespace string, overrideRBs rolloutblock.Override) error {
+	existingRBs := rolloutblock.NewOverrideFromRolloutBlocks(c.existingRolloutBlocks(namespace))
```
juliogreff: This right here makes me think `NewOverrideFromRolloutBlocks` doesn't create an override list, but just a list of rollout blocks. Maybe we should rename this.

pkg/webhook/webhook.go
juliogreff: I suppose the logic would be to check against rollout blocks if the spec (which is the only thing that triggers rollouts) was changed, and otherwise just let the changes through. That way, if one is changing labels, annotations or anything else, the webhook won't reject the update. This logic should apply to releases as well.

docs/user/rolling-out.rst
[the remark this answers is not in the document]
juliogreff: I'm of the opinion that until we need and actually do something useful with it, we should keep it out. @parhamdoustdar @icanhazbroccoli can you guys weigh in on this?

pkg/errors/rolloutblock.go
```diff
@@ -0,0 +1,45 @@
...
+func (e RolloutBlockError) Error() string {
+	return string(e)
+}
+
+func (e RolloutBlockError) ShouldRetry() bool {
+	return false
+}
+
+func (e RolloutBlockError) ShouldBroadcast() bool {
+	return false
```
juliogreff: I think that all the errors here should be broadcastable (as in, they will be also sent as an event), to be in line with the majority of our other errors.

pkg/webhook/webhook.go
```diff
@@ -175,3 +192,104 @@ func (c *Webhook) validateHandlerFunc(review *admission_v1beta1.AdmissionReview)
...
+	}
+
+	switch request.Operation {
+	case kubeclient.Create:
+		err = c.processRolloutBlocks(application.Namespace, overrideRBs)
+	case kubeclient.Update:
+		var oldApp shipper.Application
+		err = json.Unmarshal(request.OldObject.Raw, &oldApp)
+		if err != nil {
+			return nil
```
juliogreff: I think you meant `return err` here

pkg/webhook/webhook.go
```diff
@@ -175,3 +192,104 @@ func (c *Webhook) validateHandlerFunc(review *admission_v1beta1.AdmissionReview)
...
+	}
+
+	switch request.Operation {
+	case kubeclient.Create:
+		err = c.processRolloutBlocks(release.Namespace, overrideRBs)
+	case kubeclient.Update:
+		var oldRelease shipper.Release
+		err = json.Unmarshal(request.OldObject.Raw, &oldRelease)
+		if err != nil {
+			return nil
```
juliogreff: And `return err` here too

pkg/webhook/webhook.go
```diff
@@ -175,3 +192,104 @@ func (c *Webhook) validateHandlerFunc(review *admission_v1beta1.AdmissionReview)
...
+
+	nonOverriddenRBs := existingRBs.Diff(overrideRBs)
+	if len(nonOverriddenRBs) > 0 {
+		return shippererrors.NewRolloutBlockError(nonOverriddenRBs.String())
+	}
+
+	return nil
+}
+
+func (c *Webhook) existingRolloutBlocks(namespace string) []*shipper.RolloutBlock {
```
juliogreff: It seems like we're not handling any of the errors in this function. That's dangerous because it might actually allow rollouts through if they're blocked, but an error happens here.

PR #161, opened 2019-08-19
Report shipper version in User-Agent headers

juliogreff:
Closes #147

PR #164, opened 2019-08-21
Makefile and CI pipeline tweaks

juliogreff:
As requested in #161, splitting out the changes to the makefile and pipeline into its own PR.


Dockerfile.shipper
```diff
@@ -1,5 +1,5 @@
 FROM alpine:3.8
 LABEL authors="Parham Doustdar <parham.doustdar@booking.com>, Alexey Surikov <alexey.surikov@booking.com>, Igor Sutton <igor.sutton@booking.com>, Ben Tyler <benjamin.tyler@booking.com>"
```
icanhazbroccoli: @juliogreff I believe it makes sense to put your name in the list of authors and you are introducing some breaking changes to it.

Dockerfile.shipper
```diff
@@ -1,5 +1,5 @@
...
 RUN apk add ca-certificates
-ADD build/shipper /bin/shipper
+ADD build/shipper.linux-amd64 /bin/shipper
```
icanhazbroccoli: We might consider reflecting this GOARCH and GOOS in our documentation

PR #165, opened 2019-08-22
Spacing updated objects between resyncs

### LINKED ISSUE #120: Take a deeper look at our use of workqueue in the controller
While writing tests I encounter a problem in the release_controller_test: since there we also listen to events, I had a test failing because of a lot of re-processing of the same release key. 
Ignoring events in the release controller test made said tests pass, as was replacing adding items to release workqueue from `Add` to `AddRateLimited`.
To make sure, I added to application controller test an option to listen to events. When doing so, 2 tests that used to pass, now failed. Replacing `Add` with `AddRateLimited` in the application workqueue made these tests pass again.

We use [workqueue](https://godoc.org/k8s.io/client-go/util/workqueue) in our controllers.
A workqueue will not process a single item multiple times concurrently, and if an item is added multiple times before it can be processed, it will only be processed once. However, it is allowed for an item to be reenqueued while it is being processed. To add an item to a workqueue we can use `Add(item interface{})` 
We use out workqueues as RateLimiters, initialised with a DefaultControllerRateLimiter. To add an item to a RateLimitingInterface interface implementing queue, we can use `AddRateLimited(item interface{})`, adds an item to the workqueue after the rate limiter says it's ok.

hbarkal:
Addressing the spike on resyncs, adding after duration objects that on update are unchanged.

We want to space out updated objects over the time between resyncs. We want to do that only for objects that are unchanged, in order to not introduce latency to updating objects that require an update.
Closes #120


juliogreff:
One thing to keep in mind, after reading [rand.Intn](https://golang.org/pkg/math/rand/#Rand.Intn)'s documentation where it says that it panics if n <= 0: resyncs can also be 0. It's unclear how shipper would get an update with the same ResourceVersion, but I'd rather be on the safe side here.

pkg/clusterclientstore/queue.go
```diff
@@ -25,6 +26,7 @@ func (s *Store) secretWorker() {
 
 func (s *Store) bindEventHandlers() {
 	enqueueSecret := func(obj interface{}) { enqueueWorkItem(s.secretWorkqueue, obj) }
+	enqueueSecretAfterDuration := func(obj interface{}) { s.enqueueWorkItemAfterDuration(s.secretWorkqueue, obj) }
```
icanhazbroccoli: I believe there is no dedicated closure around the arguments of this function, therefore it might be simpler to call `s.enqueueWorkItemAfterDuration` with no extra syntactic sugar function wrapper: it's redefined in a few blocks below every time and is being called only once per block.

pkg/clusterclientstore/queue.go
```diff
@@ -38,8 +40,14 @@ func (s *Store) bindEventHandlers() {
 		},
 		Handler: kubecache.ResourceEventHandlerFuncs{
 			AddFunc: enqueueSecret,
-			UpdateFunc: func(_, newObj interface{}) {
-				enqueueSecret(newObj)
+			UpdateFunc: func(oldObj, newObj interface{}) {
+				oldSecret := oldObj.(*corev1.Secret)
```
icanhazbroccoli: I bet it would be great to type-cast like: oldSecret, ok := oldObj.(*corev1.Secret) to prevent a potential panic.

pkg/clusterclientstore/queue.go
```diff
@@ -61,10 +69,25 @@ func (s *Store) bindEventHandlers() {
 	})
 
 	enqueueCluster := func(obj interface{}) { enqueueWorkItem(s.clusterWorkqueue, obj) }
+	enqueueClusterAfterDuration := func(obj interface{}) { s.enqueueWorkItemAfterDuration(s.clusterWorkqueue, obj) }
```
icanhazbroccoli: Same as above: there are no benefits of having this stored in a variable

pkg/clusterclientstore/queue.go
```diff
@@ -61,10 +69,25 @@ func (s *Store) bindEventHandlers() {
...
 		AddFunc: enqueueCluster,
-		UpdateFunc: func(_, new interface{}) {
-			enqueueCluster(new)
+		UpdateFunc: func(oldObj, newObj interface{}) {
+			oldCluster, ok := oldObj.(*shipper.Cluster)
+			if !ok {
+				runtime.HandleError(fmt.Errorf("not a shipper.Cluster: %#v", oldObj))
+				return
+			}
+			newCluster, ok := newObj.(*shipper.Cluster)
```
icanhazbroccoli: Same as above: a safer typecasting would be preferable

pkg/controller/application/application_controller.go
```diff
@@ -82,8 +86,24 @@ func NewController(
...
+			}
+
+			newApp, ok := newObj.(*shipper.Application)
+			if !ok {
+				runtime.HandleError(fmt.Errorf("not a shipper.Application: %#v", newObj))
+				return
+			}
+
+			if oldApp.GetResourceVersion() == newApp.GetResourceVersion() {
+				c.enqueueAppAfterDuration(newObj)
```
icanhazbroccoli: The name of this method is a bit misleading as it provides no access to the specific timeout. I.e.: calling it like: `c.enqueueAppAfterDuration(newObj, time.Second)` would match with my expectations from a method called like this. The method name should give a signal the caller doesn't really have a control over it. Proposed name: `enqueueAppLater(newObj)`. Reading `enqueueAppAfterDuration` code indicates the delay would be chosen based on the resync period. This raises a question: is it a permanent strategy to choose the juffy based on resync period or it's just something that makes sense by now but it's a subject of future reconsideration? I'm mainly asking because initially I saw this code and wanted to propose a method rename: `enqueueAppOnResync`, which eventually made very little sense to me as a resync mechanism will still bting it to the resync stage and therefore might be dropped completely.

pkg/controller/application/application_controller.go
```diff
@@ -215,6 +235,17 @@ func (c *Controller) enqueueApp(obj interface{}) {
...
 }
 
+func (c *Controller) enqueueAppAfterDuration(obj interface{}) {
+	key, err := cache.MetaNamespaceKeyFunc(obj)
+	if err != nil {
+		runtime.HandleError(err)
+		return
+	}
+
+	duration := shippercontroller.GetDuration(c.resyncPeriod)
```
icanhazbroccoli: This is the duration I meant above: if it's going to stay like this for a longer period of time, we can easily ignore same-UUID notifications. Another danger I see here is: it's an ever-growing queue which can accumulate itself within shipper runtime and therefore cause resource leakage.

pkg/controller/application/application_controller_test.go
```diff
@@ -891,8 +891,8 @@ func (f *fixture) newController() (*Controller, shipperinformers.SharedInformerF
 
 	const noResyncPeriod time.Duration = 0
 	shipperInformerFactory := shipperinformers.NewSharedInformerFactory(f.client, noResyncPeriod)
-
-	c := NewController(f.client, shipperInformerFactory, f.resolveChartVersion, record.NewFakeRecorder(42))
+	var resync time.Duration = 1
```
icanhazbroccoli: Please consider using constants from time package, say, time.Second instead of simple casted integer: it provides time-duration semantics to the reader.

pkg/controller/capacity/capacity_controller_test.go
```diff
@@ -671,7 +672,7 @@ func (f *fixture) newController() *Controller {
 func (f *fixture) runInternal() *Controller {
 	f.initializeFixture()
 
-	controller := f.newController()
+	controller := f.newController(time.Duration(1))
```
icanhazbroccoli: s/time.Duration(1)/time.Second/

pkg/clusterclientstore/queue.go
hbarkal: It is being redefined but with a different workqueue. I can make more readable and unwrap it...

pkg/clusterclientstore/queue.go
hbarkal: I thought that was the job of the filter? If it's not, I'll fix that

pkg/clusterclientstore/queue.go
```diff
@@ -37,9 +37,23 @@ func (s *Store) bindEventHandlers() {
...
+				}
+				newSecret, ok := newObj.(*corev1.Secret)
+				if !ok {
+					runtime.HandleError(fmt.Errorf("not a Secret: %#v", newObj))
+					return
+				}
+				if oldSecret.GetResourceVersion() == newSecret.GetResourceVersion() {
+					s.enqueueSecretLater(newObj)
+				} else {
+					s.enqueueSecret(newObj)
```
isutton: `enqueueSecret` is `enqueueSecretLater` with `0` duration.

pkg/clusterclientstore/queue.go
```diff
@@ -80,19 +107,51 @@ func (s *Store) bindEventHandlers() {
...
+	key, err := kubecache.DeletionHandlingMetaNamespaceKeyFunc(obj)
+	if err != nil {
+		runtime.HandleError(err)
+		return
+	}
+
+	s.secretWorkqueue.Add(key)
+}
+
+func (s *Store) enqueueSecretLater(obj interface{}) {
```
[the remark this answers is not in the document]
isutton: The same applies for `enqueue*Later`.

pkg/controller/capacity/capacity_controller.go
```diff
@@ -289,14 +311,18 @@ func (c *Controller) capacityTargetSyncHandler(key string) error {
...
 		runtime.HandleError(err)
 		return
 	}
 
-	c.capacityTargetWorkqueue.Add(key)
+	if duration > 0 {
+		c.capacityTargetWorkqueue.AddAfter(key, duration)
+	} else {
+		c.capacityTargetWorkqueue.Add(key)
+	}
```
isutton: Is this check required in here? I'd think that `Add(key)` and `AddAfter(key, 0)` to be equivalent, and any optimization like this to be performed at the queue implementation level.

pkg/controller/rolloutblock/rolloutblock_controller.go
```diff
@@ -141,9 +145,17 @@ func (c *Controller) onUpdateRelease(oldObj interface{}, newObj interface{}) {
 	// rolloutblock object should be updated if a release has added it to override annotation,
 	// or removed it from override annotation
 	if len(oldOverrideRBs.Diff(newOverrideRBs)) > 0 {
-		c.enqueueReleaseBlock(oldObj)
+		if oldRel.GetResourceVersion() == newRel.GetResourceVersion() {
+			c.enqueueRolloutBlockKeysLater(oldOverrideRBs)
```
isutton: Same principle of `*After` would apply in here (unless `*Later` is equivalent to `*After` with some configuration deep in the controller). If not, adding a `time.Duration` argument here also makes sense, and rename it to `...After`.

After reviewed the `enqueueRolloutBlockKeysLater()` implementation, it'd still make sense to pass a default duration to this method, allowing the caller to have control on the default computation of the duration.

juliogreff:
As we intend to merge #209 and #213 into the v0.7 release, I suppose this makes this PR no longer necessary, especially now that it has a bunch of conflicts to be resolved :/ So I'm closing this in favor of the mentioned PRs.

PR #166, opened 2019-08-27
ReleaseController application picks the tip of the release chain as contender

icanhazbroccoli:
The current implementation has a slightly relaxed algorithm of finding
contender and incumbent in the chain of releases: in particular, the
latest (sorted by generation) scheduled release is identified as a
contender and the latest complete one is our incumbent.

This commit changes the approach and forces release controller to start
using releaseutil methods for finding incumbent and contender. The
incumbent finding is still very naive as it's relying on release status
to figure out whether it's complete or not. On the improvement side, we
got rid of using the most dangerous condition in release object:
scheduled criteria. From now on, we always pick the tip of the release
chain.

### DIFF
--- a/pkg/chart/repo/repo.go
+++ b/pkg/chart/repo/repo.go
@@ -224,7 +224,11 @@ func (r *Repo) FetchRemote(cv *repo.ChartVersion) (*chart.Chart, error) {
 	url := chartURL.String()
 	data, err := r.fetcher(url)
 	if err != nil {
-		return nil, err
+		chart, convErr := newChart(cv)
+		if convErr != nil {
+			return nil, shippererrors.NewChartRepoInternalError(convErr)
+		}
+		return nil, shippererrors.NewChartFetchFailureError(chart, err)
 	}
 
 	chart, err := loadChartData(data)
@@ -306,4 +310,15 @@ func chart2file(cv *repo.ChartVersion) string {
 	version = strings.Replace(version, "/", "-", -1)
 
 	return fmt.Sprintf("%s-%s.tgz", name, version)
+}
+
+func newChart(cv *repo.ChartVersion) (*shipper.Chart, error) {
+	if len(cv.URLs) < 1 {
+		return nil, fmt.Errorf("chart version is missing URLs")
+	}
+	return &shipper.Chart{
+		Name:    cv.GetName(),
+		Version: cv.GetVersion(),
+		RepoURL: cv.URLs[0],
+	}, nil
 }
--- a/pkg/controller/release/release_controller.go
+++ b/pkg/controller/release/release_controller.go
@@ -2,6 +2,7 @@ package release
 
 import (
 	"fmt"
+	"reflect"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
@@ -19,6 +20,7 @@ import (
 	shipperclient "github.com/bookingcom/shipper/pkg/client/clientset/versioned"
 	shipperinformers "github.com/bookingcom/shipper/pkg/client/informers/externalversions"
 	shipperlisters "github.com/bookingcom/shipper/pkg/client/listers/shipper/v1alpha1"
+	"github.com/bookingcom/shipper/pkg/controller"
 	shippererrors "github.com/bookingcom/shipper/pkg/errors"
 	releaseutil "github.com/bookingcom/shipper/pkg/util/release"
 )
@@ -304,14 +306,13 @@ func (c *Controller) syncOneReleaseHandler(key string) error {
 			WithShipperKind("Release")
 	}
 
-	rel = rel.DeepCopy()
-
 	if releaseutil.HasEmptyEnvironment(rel) {
 		return nil
 	}
 
-	klog.V(4).Infof("Start processing Release %q", key)
+	var initialRel *shipper.Release = rel.DeepCopy()
 
+	klog.V(4).Infof("Start processing Release %q", key)
 	scheduler := NewScheduler(
 		c.clientset,
 		c.clusterLister,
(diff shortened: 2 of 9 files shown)


icanhazbroccoli:
@kanatohodets @juliogreff Hey folks! Mind dropping a line on this one as we all had concerns regarding this change, it would be great to keep it tracked. Thanks!

pkg/util/application/releases.go
```diff
@@ -21,7 +21,14 @@ func GetContender(appName string, rels []*shipper.Release) (*shipper.Release, er
 // An incumbent release is the first release in this slice that is considered
 // completed.
 func GetIncumbent(appName string, rels []*shipper.Release) (*shipper.Release, error) {
+	contender, err := GetContender(appName, rels)
+	if err != nil {
+		return nil, err
+	}
 	for _, r := range rels {
+		if r == contender {
```
ksurent: This compares two pointers. I don't think the API makes any guarantees about the safety of doing so. Better to compare by name.

isutton: `GetContender` returns the pointer to the contender if present in `rels`, so this should be safe.

https://github.com/bookingcom/shipper/blob/3c1a7d519962b94fea405ab5df1367c5700e28ce/pkg/util/application/releases.go#L18-L30

ksurent: The API makes no guarantees about the safety of that. It's working today but tomorrow it might return a shallow copy. Either explicitly document the guarantee or compare by name (better).

pkg/controller/release/release_controller.go
```diff
@@ -313,19 +313,12 @@ func (c *Controller) syncOneReleaseHandler(key string) error {
...
-	// sequentual updates on the same object experienced apiserver
-	// rejections due to the passed object outdate state. This happened due
-	// to the synchronisation/replication lag in etcd. This approach helps
-	// to eliminate this as a notification will be delivered once all
-	// parties are in sync.
 	if !releaseHasClusters(rel) {
-		_, err = scheduler.ChooseClusters(rel, false)
-	} else {
+		shouldForce := false
+		_, err = scheduler.ChooseClusters(rel, shouldForce)
```
juliogreff: If you don't use the returned release here, the subsequent call to ScheduleRelease will always fail due to a conflict, as the resource versions of what you have in memory and what's in the apiserver will be different. That'd be equivalent to the previous version of the code.

pkg/util/application/releases.go
isutton: Can you point where this is stated? I'd love to learn more about that, as I thought the assumption is one shouldn't update an object obtained through the API, but not regarding the value themselves.

juliogreff:
My original worry on this change was actually unfounded (I took a [code comment to its word](https://github.com/bookingcom/shipper/blob/4b60a99/pkg/controller/application/application_controller.go#L515-L523), but that [has since been addressed](https://github.com/bookingcom/shipper/pull/172/commits/ffd0a3642560b1e57678bb01899cc8bb134db656)). I'm actually quite comfortable with the spirit of the change now that my misunderstanding has been cleared up.

There is a quite unlikely scenario that might be worth considering though. Imagine that a healthy release `A` is being replaced by release `B`. If `B` has its target step advanced all the way to the last step in the strategy, but a new release `C` gets created before `B` has time to complete (say, because of an emergency hotfix), the incumbent will remain being `A`, even though it might've lost quite some capacity during the rollout, and _maybe_ in this particular case `B` could've been a better candidate.

Honestly I can't think of a good way around this (maybe preventing new releases when there's an incomplete contender?), so this is just putting it out there. As I said earlier, the idea behind the current state of this PR is sound to me now.

pkg/controller/release/release_controller.go
```diff
@@ -313,22 +315,23 @@ func (c *Controller) syncOneReleaseHandler(key string) error {
...
+			rel,
+			corev1.EventTypeNormal,
+			"ClustersSelected",
+			"Set clusters for %q to %v",
+			controller.MetaKey(rel),
+			rel.Annotations[shipper.ReleaseClustersAnnotation],
+		)
 	}
 
+	rel, err = scheduler.ScheduleRelease(rel.DeepCopy())
```
juliogreff: Is there reason to deep copy the release here again?

juliogreff: Unrelated to the previous question, now that choosing clusters no longer saves the release, what do you think about moving the `.Update()` call here to the controller instead of buried inside the scheduler, so both the happy path and the error path are handled in the same place?

pkg/controller/release/release_controller.go
```diff
@@ -317,7 +318,8 @@ func (c *Controller) syncOneReleaseHandler(key string) error {
 
 	if !releaseHasClusters(rel) {
 		shouldForce := false
-		if err := c.chooseClusters(rel, shouldForce); err != nil {
+		rel, err = c.chooseClusters(rel, shouldForce)
```
isutton: Does any usage of `rel` up to this point require it to be deeply copied? Perhaps using `curRel` deep copy in this scope only as input of `chooseClusters` and keep referencing `curRel` everywhere can make it less ambiguous?

Another option would be `chooseClusters` perform a deep copy itself.

pkg/controller/release/release_controller.go
```diff
@@ -350,11 +350,16 @@ func (c *Controller) syncOneReleaseHandler(key string) error {
 			err.Error(),
 		)
 		releaseutil.SetReleaseCondition(&rel.Status, *condition)
+	}
 
-		if _, updateErr := c.clientset.ShipperV1alpha1().Releases(namespace).Update(rel); updateErr != nil {
-			return shippererrors.NewKubeclientUpdateError(rel, updateErr)
+	// any better ideas?
+	if !reflect.DeepEqual(curRel, rel) {
```
isutton: If `rel` is created in the block above, then `chooseClusters` could be used to detect whether or not the update should happen.

pkg/controller/release/release_controller.go
```diff
@@ -350,11 +350,16 @@ func (c *Controller) syncOneReleaseHandler(key string) error {
...
+		if _, err := c.clientset.ShipperV1alpha1().Releases(namespace).Update(rel); err != nil {
+			return shippererrors.NewKubeclientUpdateError(rel, err)
 		}
+	}
 
+	if err != nil {
```
isutton: What is this guarding against?

pkg/controller/release/release_controller.go
```diff
@@ -559,26 +564,26 @@ func reasonForReleaseCondition(err error) string {
...
 			shipper.SchemeGroupVersion.WithKind("Cluster"),
 			"", selector, err)
 	}
 
 	selectedClusters, err := computeTargetClusters(rel, allClusters)
 	if err != nil {
-		return err
+		return nil, err
 	}
 	setReleaseClusters(rel, selectedClusters)
```
isutton: `chooseClusters` could return `nil, nil` here to indicate that no change has been applied.

pkg/util/application/releases.go
icanhazbroccoli: Comparison is based on name equality, not pointers: https://github.com/bookingcom/shipper/pull/166/files#diff-80b2e724ff7817e14b9de5d4d3a25c9cR30

pkg/controller/release/release_controller.go
icanhazbroccoli: Fixed, thanks!

pkg/controller/release/release_controller.go
icanhazbroccoli: Moved, tested :-)

pkg/controller/release/release_controller.go
icanhazbroccoli: It's a but tricky here: there is only 1 Update statement and a few code ways that can modify the rel object:
* ChooseClusters chooses relevant placement and puts this data into annotations
* scheduling succeeds and sets Scheduled condition to true
* scheduling fails and sets Scheduled condition to false
All 3 paths lead to a single Update point which triggers if there is effectively any difference between rel and curRel. From this perspective there are 2 components (effectively 3) that should indicate whether a change has been made.

pkg/controller/release/release_controller.go
icanhazbroccoli: Renamed this err to scheduleErr. The idea is to get control back from ScheduleRelease routine, check errors and if there are some, update corresponding conditions; then check if there are any effective changes to the release after all these cycles. Afterwards it checks if there was an error from ScheduleRelease and if there was, exit the routine.

pkg/chart/repo/repo.go
```diff
@@ -307,3 +314,14 @@ func chart2file(cv *repo.ChartVersion) string {
 
 	return fmt.Sprintf("%s-%s.tgz", name, version)
 }
+
+func chartVerToChartSpec(cv *repo.ChartVersion) (*shipper.Chart, error) {
```
isutton: This would be better named `newChart`, since this is a constructor.

pkg/chart/repo/repo.go
```diff
@@ -224,7 +224,14 @@ func (r *Repo) FetchRemote(cv *repo.ChartVersion) (*chart.Chart, error) {
 	url := chartURL.String()
 	data, err := r.fetcher(url)
 	if err != nil {
-		return nil, err
+		chartspec, convErr := chartVerToChartSpec(cv)
```
isutton: Renaming `chartVerToChartSpec` to `newChart` makes more sense in here, and renaming the variables as well:

```go
chart, newErr := newChart(cv)
```

pkg/controller/release/release_controller.go
```diff
@@ -323,46 +324,51 @@ func (c *Controller) syncOneReleaseHandler(key string) error {
...
 			corev1.ConditionFalse,
 			reason,
-			err.Error(),
+			scheduleErr.Error(),
 		)
-		releaseutil.SetReleaseCondition(&rel.Status, *condition)
-
-		if _, updateErr := c.clientset.ShipperV1alpha1().Releases(namespace).Update(rel); updateErr != nil {
-			return shippererrors.NewKubeclientUpdateError(rel, updateErr)
+		releaseutil.SetReleaseCondition(&initialRel.Status, *condition)
```
juliogreff: It's a bit surprising to me that we refer to `initialRel` anywhere other than the DeepEqual check a bit later in this func. Is this for a special reason?

pkg/controller/release/release_controller_test.go
```diff
@@ -927,7 +939,7 @@ func (f *fixture) expectAssociatedObjectsCreated(release *shipper.Release, clust
 			"Normal ReleaseScheduled Created CapacityTarget \"%s\"",
 			relKey,
 		),
-	}
+	}...)
```
juliogreff: Bit of a nitpick, but the ellipsis here are a bit weird, as we're building a `[]string` just to unpack it immediately.

pkg/controller/release/release_controller.go
juliogreff: Nevermind, after reading further I see why, things operating on the release return `nil` on error. Makes sense now.

PR #167, opened 2019-08-27
docs: value of enable-helm-release-workaround requires quotes

juliogreff:
This has wasted one hour of life of at least one person: all labels must
be strings, but the YAML parser serializes `true` (no quotes) as a
boolean. This makes the apiserver barf with gibberish deserialization
errors when trying to apply the object.


isutton:
> but the YAML parser serializes `true` (no quotes) as a boolean

This is the reason Kubernetes uses `True`, `False` and `Unknown` as values for some fields. 

I understand this work unit is related to documentation, but perhaps this could lead to some scheme to enable experimental features on the controller itself?

```yaml
annotations:
    shipper.booking.com/features: helm-release-workaround, chart-range
```

PR #175, opened 2019-09-04
Turn shipper_objects_release_durations into a histogram

juliogreff:
Back in 24787c85dea459e07ca16e6b43cccaed7c5ea1c7, we decided to collect
a summary instead of a histogram because we did not know how the
distribution of values looked like. It turns out that with a busy enough
cluster, a lot of releases may remain incomplete for... well, forever.
And although we know that, we still don't know any more about the
distribution of releases that actually eventually complete.

In any case, this distribution might be different for every shipper
installation, so there's not a whole lot of sense in hardcoding these
values anyway. So shipper-state-metrics learned a new argument,
`-release-duration-buckets`, where you can pass a comma-separated list
of buckets, in seconds, for the histogram.


cmd/shipper-state-metrics/main.go
```diff
@@ -21,11 +24,12 @@ import (
 )
 
 var (
-	masterURL    = flag.String("master", "", "The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.")
-	kubeconfig   = flag.String("kubeconfig", "", "Path to a kubeconfig. Only required if out-of-cluster.")
-	resyncPeriod = flag.String("resync", "5m", "Informer's cache re-sync in Go's duration format.")
```
icanhazbroccoli: This flag should be dropped from shipper deployment object, otherwise it will fail to start: https://github.com/bookingcom/shipper/blob/master/kubernetes/shipper.deployment.yaml#L28

juliogreff: That deployment is for the `shipper` cmd. `shipper-state-metrics.deployment.yaml` doesn't pass any arguments.

icanhazbroccoli: You are absolutely right.

PR #181, opened 2019-09-13
crds: additional printer columns for Application

juliogreff:
Now, when doing `kubectl get app`, output will look like the following:

```
NAME        LATEST RELEASE         ROLLING OUT   AGE
snowflake   snowflake-306ba2a4-0   False         22h
```

### DIFF
--- a/crd/Application-crd.yaml
+++ b/crd/Application-crd.yaml
@@ -4,6 +4,19 @@ metadata:
   # name must match the spec fields below, and be in the form: <plural>.<group>
   name: applications.shipper.booking.com
 spec:
+  additionalPrinterColumns:
+  - JSONPath: .status.history[-1]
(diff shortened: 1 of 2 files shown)


crd/Application-crd.yaml
```diff
@@ -4,6 +4,19 @@ metadata:
   # name must match the spec fields below, and be in the form: <plural>.<group>
   name: applications.shipper.booking.com
 spec:
+  additionalPrinterColumns:
+  - JSONPath: .status.history[-1]
```
icanhazbroccoli: Proposed change:
```
- JSONPath: .status.history[(@.length-1)]
```

juliogreff: Will do! Thanks!

PR #183, opened 2019-09-13
installer: read ReleaseLabel if InstallationTargetOwnerLabel is missing

juliogreff:
If an object was created before we introduced self-contained
InstallationTargets, it will not contain InstallationTargetOwnerLabel,
values, so it will need to be migrated. We do so by getting the older
ReleaseLabel.

This fixes an issue where previously installed objects wouldn't be
touched during a new release, erroring out with a
`MissingInstallationTargetOwnerLabelError`.

### DIFF
--- a/pkg/controller/installation/installer.go
+++ b/pkg/controller/installation/installer.go
@@ -494,9 +494,19 @@ func (i *Installer) installManifests(
 			continue
 		}
 
-		owner, ok := existingObj.GetLabels()[shipper.InstallationTargetOwnerLabel]
+		labels := existingObj.GetLabels()
+		owner, ok := labels[shipper.InstallationTargetOwnerLabel]
 		if !ok {
-			return shippererrors.NewMissingInstallationTargetOwnerLabelError(existingObj)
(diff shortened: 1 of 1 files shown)


pkg/controller/installation/installer.go
```diff
@@ -494,9 +494,19 @@ func (i *Installer) installManifests(
...
-		owner, ok := existingObj.GetLabels()[shipper.InstallationTargetOwnerLabel]
+		labels := existingObj.GetLabels()
+		owner, ok := labels[shipper.InstallationTargetOwnerLabel]
 		if !ok {
-			return shippererrors.NewMissingInstallationTargetOwnerLabelError(existingObj)
+			// If an object was created before we introduced
+			// self-contained InstallationTargets, it will not
+			// contain InstallationTargetOwnerLabel, values, so it
+			// will need to be migrated. We do so by getting the
+			// older ReleaseLabel.
```
icanhazbroccoli: Thanks a lot for a descriptive comment here!

PR #186, opened 2019-09-13
Update chart.go

### LINKED ISSUE #145: Render *.yml templates inside charts
Shipper suffers from a similar issue as helm did at some point: https://github.com/helm/helm/pull/3646

Probably as an oversight, we only try to render *.yaml files, and completely ignore *.yml. We should treat them both as being proper YAML.

Relevant code: https://github.com/bookingcom/shipper/blob/master/pkg/chart/chart.go#L61-L63

ilmanzo:
fix  #145


icanhazbroccoli:
Yay! Thanks folks!

PR #196, opened 2019-09-23
release controller: introduce a Blocked condition

juliogreff:
release controller: introduce a Blocked condition

Having to check whether rollouts are blocked from inside
`ScheduleRelease` ends up with releases going unscheduled, which is not
really true: they just can't proceed, whatever was scheduled previously
remains scheduled.

It also came to our attention that the application work queue in the
release controller was not honoring rollout blocks at all. This wasn't
noticed because it'd take a shipper restart, or a resync to be
triggered, as normaly the application work queue only receives work when
the release work queue can schedule a release successfully. That's now
fixed.

Additionally, we now have a better wrapper method over what's being
changed in a sync step, and we're better at not issuing new updates when
nothing has changed about the object.


pkg/controller/release/release_controller.go
```diff
@@ -324,12 +354,37 @@ func (c *Controller) syncOneReleaseHandler(key string) error {
 		c.recorder,
 	)
 
+	initialRel := rel.DeepCopy()
+
+	canSchedule, err := scheduler.CanScheduleRelease(rel)
+	if canSchedule {
```
isutton: Won't `canSchedule` be false if an error is also return?

pkg/controller/application/application_controller.go
```diff
@@ -396,19 +394,30 @@ func (c *Controller) processApplication(app *shipper.Application) error {
 		}
 	}
 
-	// existing rollout blocks in the system
-	var nsRolloutBlocks, globalRolloutBlocks []*shipper.RolloutBlock
-	if nsRolloutBlocks, err = c.rbLister.RolloutBlocks(app.Namespace).List(labels.Everything()); err != nil {
-		klog.Warningf("failed to list Namespace RolloutBlocks %s", err.Error())
+	rolloutBlocked, events, err := rolloutblock.BlocksRollout(c.rbLister, app)
+	for _, ev := range events {
+		c.recorder.Event(app, ev.Type, ev.Reason, ev.Message)
```
icanhazbroccoli: Do you think it might make sense to pass the instance of recorder inside `rolloutblock.BlocksRollout`?

pkg/controller/release/application.go
```diff
@@ -93,6 +94,15 @@ func (c *Controller) syncOneApplicationHandler(key string) error {
 		return err
 	}
 
+	// If rollouts are blocked for the contender release, here we will just
+	// bail out without updating any conditions. They've already been set
+	// bot for the Application in the application controller, and for the
+	// release in the main queue of the release controller.
+	rolloutBlocked, _, err := rolloutblock.BlocksRollout(c.rolloutBlockLister, contender)
+	if rolloutBlocked {
+		return err
```
icanhazbroccoli: Not sure if it concerns us: shall we check if err is non-nil?

juliogreff:
@isutton @parhamdoustdar @icanhazbroccoli I agree that the boolean and the error being equivalent (at least right now) reads a little bit weird, but I couldn't come up with something more idiomatic. There's also a minor issue in that a rollout block there isn't actually an error, I went that way just for the convenience of reusing the errors we already had. I'll address this with a new PR, not to introduce even more changes to 0.6, this is already on the large side as it is.

In any case, you guys raise a valid point that we should still check for `err != nil` as well, to future proof changes to the assumption the error and boolean are equivalent. I'll push those changes shortly.

PR #203, opened 2019-09-24
update kubernetes objects for webhooks in shipperctl

### LINKED ISSUE #162: shipperctl does not update kubernetes objects for webhooks
shipperctl needs to create several kubernetes objects when setting up shipper clusters. On a first try, everything works alright, but a subsequent run of shipperctl will not update the webhook service nor webhook configuration. This might bring issues in the future for users upgrading from one version of shipper to the next, if they happen to change any of these objects.

hbarkal:
shipperctl checks if ValidatingWebhookService and ValidatingWebhookConfiguration exists, and if so, it does not update these objects.

Following ValidatingWebhookSecret, CRD and Cluster create or update logic, shipperctl should create the service and the webhook configuration if they don't exists. If they do, shipperctl should update them.

Closes #162


cmd/shipperctl/configurator/cluster.go
```diff
@@ -483,6 +490,18 @@ func (c *Cluster) CreateValidatingWebhookService(namespace string) error {
...
+			_, err = c.KubeClient.CoreV1().Services(namespace).Create(service)
+			return err
+		} else {
+			return err
+		}
+	}
+
+	existingSerivce.Spec.Selector = service.Spec.Selector
+	existingSerivce.Spec.Ports = service.Spec.Ports
+	_, err = c.KubeClient.CoreV1().Services(namespace).Update(existingSerivce)
```
icanhazbroccoli: Any chance you could write a unit test to cover this Update part please?

PR #209, opened 2019-09-30
Introduce JitteredRateLimiter and retry errors indefinitely

### LINKED ISSUE #207: Controllers should retry retriable errors indefinitely
Currently, if we run shipper without resyncs, retriable errors have a cap on the number of times they'll be retried. That's not a problem if they're not really a transient issue, but if they are, and need more retries than the 11 (?) we have currently configured for most controllers, they'll be dropped from queues and never looked at again. Ideally, we'd just keep on retrying them forever, at reasonable intervals (so not too fast, maybe a few seconds), and some jitter, so transient errors to a lot of objects don't trigger thundering herds.

juliogreff:
Shipper suffers a little bit with thundering herds when retrying errors,
every time it starts, and during outages of any of its dependencies, as
all objects that errored out are reprocessed all at the same time.

To avoid that issue, we're introducing a rate limiter that adds a little
bit of jitter to each requeue. We also make all controllers use this new
JitteredRateLimiter, and they no longer drop items from the queue after
a certain amount of retries.

This closes #207.


pkg/controller/janitor/janitor_controller.go
```diff
@@ -58,7 +52,7 @@ func NewController(
 
 	controller := &Controller{
 		recorder:           recorder,
-		workqueue:          workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter()),
+		workqueue:          workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "janitor_controller_installationtargets"),
```
icanhazbroccoli: Mind dropping a line why we favour workqueue limiter to shipperworkqueue here?

juliogreff: Actually, just my mistake. Good catch ^_^

PR #215, opened 2019-10-04
shipperctl: introduce `shipperctl chart render`

juliogreff:
It's useful to know what kind of objects shipper is rendering before
they reach the api server. For debugging issues with YAML
serializing/deserializing, for instance.

For that, you can now run `shipperctl chart render --app foobar` or
`shipperctl chart render --relelase foobar-deadbeef-0`, and shipperctl
will print a whole bundle of YAML to stdout, with each object separated
by a `---`.

This is currently only useful for applications that already exist, as
you can't manually pass standalone chart spec or values. We'll
eventually support that once we figure out a non-awkard interface for
doing that.


cmd/shipperctl/cmd/chart/command.go
```diff
@@ -0,0 +1,142 @@
...
+		c.ReleaseName,
+		c.Namespace,
+		c.ChartValues,
+	)
+	if err != nil {
+		return err
+	}
+
+	for _, v := range rendered {
+		fmt.Printf("%s\n---\n", v)
```
icanhazbroccoli: Consider printing to a buffer and returning a rendered string rather than a direct flush to STDOUT. The control flow will gain more ownership over the rendered data.

cmd/shipperctl/cmd/chart/command.go
```diff
@@ -0,0 +1,142 @@
...
+	} else if appName != "" {
+		app, err := shipperclient.Applications(namespace).Get(appName, getoptions)
+		if err != nil {
+			return c, err
+		}
+
+		c.ReleaseName = fmt.Sprintf("%s-%s-%d", appName, "foobar", 0)
+		c.ChartSpec = app.Spec.Template.Chart
+		c.ChartValues = app.Spec.Template.Values
+	}
```
icanhazbroccoli: Shall we break it down to separate routines `renderRelease` and `renderApp`?

cmd/shipperctl/cmd/chart/command.go
juliogreff: I'd agree if we eventually move parts of this to a library kind of package, but `renderChart` here is the top level func executed by the command, it's got nowhere to return data to.

cmd/shipperctl/cmd/chart/command.go
juliogreff: This code just gathers a chart spec + values, so they don't render anything. I imagine you mean that the UI should have two separate commands for rendering a release or app?

icanhazbroccoli: Yes, this is very much what I meant.

PR #216, opened 2019-10-04
chart: split multiple objects inside the same yaml file

juliogreff:
An user reported to us that Shipper is not rendering multiple objects
inside the same yaml file (as in, multiple segments separated by a
`---`) in a chart. For instance, the following YAML excerpt will only
result in the first object being rendered, while the second one is
completely ignored:

```
apiVersion: v1
kind: Service
---
apiVersion: v1
kind: Pod
```

This patch is pretty much lifted from Helm itself:
https://github.com/helm/helm/blob/2d77db1/pkg/releaseutil/manifest.go#L37-L59

### DIFF
--- a/pkg/chart/chart.go
+++ b/pkg/chart/chart.go
@@ -2,6 +2,7 @@ package chart
 
 import (
 	"fmt"
+	"regexp"
 	"sort"
 	"strings"
 
@@ -53,15 +54,7 @@ func Render(chart *helmchart.Chart, name, ns string, shipperValues *shipper.Char
 		return nil, fmt.Errorf("could not render the chart: %s", err)
 	}
 
-	objects := make([]string, 0, len(rendered))
-	for n, o := range rendered {
-		// It's annoying to ensure that your template doesn't render spurious
-		// newlines.
-		o = strings.TrimSpace(o)
-		if len(o) > 0 && (strings.HasSuffix(n, ".yaml") || strings.HasSuffix(n, ".yml")) {
-			objects = append(objects, o)
-		}
-	}
+	objects := CollectObjects(rendered)
 
 	ks, err := newKindSorter(objects, InstallOrder)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/chart/chart.go
```diff
@@ -71,3 +64,28 @@ func Render(chart *helmchart.Chart, name, ns string, shipperValues *shipper.Char
...
+	objects := make([]string, 0, len(rendered))
+
+	for n, o := range rendered {
+		if !strings.HasSuffix(n, ".yaml") && !strings.HasSuffix(n, ".yml") {
+			continue
+		}
+
+		// Making sure that any extra whitespace in YAML stream doesn't
+		// interfere in splitting documents correctly.
+		objs := sep.Split(strings.TrimSpace(o), -1)
```
zoidbergwill: Maybe my suggestion is objectively worse, but we could convert from string to YAML then loop through the documents in the file and do YAML to string instead.

I guess it's relatively sane to assume no one will have something that matches that regex anywhere else in there YAML, though I'm sure you could, because YAML is the wild west.

icanhazbroccoli: @zoidbergwill thanks for this comment. It made me think we're doing something weird here as golang yaml parser supports streaming decoder. Here is a code sample i made to illustrate this: https://play.golang.org/p/Nz-uG2SFtpR. @juliogreff Do you think we can adopt a similar approach here?

juliogreff: We do not parse YAML manifests ourselves at any point right now, so we're not the ones doing weird things directly, the issue is that we use a [UniversalDeserializer](https://godoc.org/k8s.io/apimachinery/pkg/runtime/serializer#CodecFactory.UniversalDecoder) in a couple of places, and that deserializes `[]byte` into `runtime.Object`. It deserializes things into a *single* `runtime.Object`, no matter how many objects are in the blob.

I guess we _could_ ditch that, take matters into our own hands and parse things ourselves effectively making `shipperchart.Render` return `[]runtime.Object` instead of `[]string`. I don't know how much complexity would hide behind that. One thing that we'd lose for sure, though, is the ability to render JSON files inside the charts, as the `UniversalDecoder` supports that out of the box, [even though shipper doesn't](https://github.com/bookingcom/shipper/blob/d10f43b/pkg/chart/chart.go#L61-L63).

I'd be up for giving the yaml parsing a try, but only if we completely replace the UniversalDecoder. I'm not a fan of that because of the functionality we'd lose and the complexity we'd adopt, but let's discuss :) I wouldn't try parsing the YAML here to split the manifests, to turn it back into YAML blobs to hand the UniversalDecoder, without doing some benchmarking to see what's the performance impact, as shipper renders charts quite a lot currently.

icanhazbroccoli: Thanks a lot for your explanation @juliogreff, it totally makes sense now. Indeed, the roads are winding here. It makes me think my suggestion is still not crazy silly if, say, we propose something like:
```go
type StreamingDecoder interface {    
    Decode(reader io.Reader, defaults *schema.GroupVersionKind, into Object) (Object, *schema.GroupVersionKind, error)
}
```
If this was the case, parsing of multi-frame documents would be easier in a situation like that.

icanhazbroccoli: @juliogreff as a followup on my previous comment: I understand I'm not helping here :-) This is my way of saying: I agree with your original implementation.

PR #217, opened 2019-10-08
A condition state transition emits a tracing event

icanhazbroccoli:
Based on the internal user experience, it was concluded that Shipper's
strategy of event reporting is clumsy: once a fact was reported in
events, it's unclear whether it's still the case or it was disproved by
some later coming events. In particular: a user reported uncertainty on
their application status when Shipper reported a chart fetch problem in
events and once the fetch problem was gone, nothing indicated if it was
really gone. Events, unlike conditions, are mainly for tracing the
history of changes rather than checking the current status. Therefore we
started seeking for a better application of events.

This commit proposes to use events for high-level object condition
change tracking: every time a condition changes it's value, controller
leaves a record in the relevant object event journal. We are focusing on
application controller and application objects for now for the sake of
reducing the scope of the discussion. If the general course seems legit,
we can extend the approach to other controllers and objects.

`apputil.ApplicationConditionDiff` is a subject of a deep internal rework.

A very early example of application object event journal:
```
...
Events:
  Type     Reason                       Age                 From                    Message
  ----     ------                       ----                ----                    -------
  Normal   ApplicationConditionChanged  40m                 application-controller  [<nil>] -> [RollingOut], [<nil>] -> [False], [<nil>] -> [ChartVersionResolutionFailed], [<nil>] -> [failed to resolve chart version [name: "icanhazbroccoli-test", version: "0.0.2", repo: "http://0.0.0.0:8879"]: no chart version found]
  Warning  FailedApplication            40m (x16 over 40m)  application-controller  failed to resolve chart version [name: "icanhazbroccoli-test", version: "0.0.2", repo: "http://0.0.0.0:8879"]: no chart version found
  Normal   ApplicationConditionChanged  40m                 application-controller  [<nil>] -> [Blocked], [<nil>] -> [False], [<nil>] -> [], [<nil>] -> []
  Normal   ApplicationConditionChanged  40m                 application-controller  [<nil>] -> [Aborting], [<nil>] -> [False], [<nil>] -> [], [<nil>] -> []
  Normal   ApplicationConditionChanged  40m                 application-controller  [<nil>] -> [ValidHistory], [<nil>] -> [True], [<nil>] -> [], [<nil>] -> []
  Normal   ApplicationConditionChanged  40m                 application-controller  [<nil>] -> [ReleaseSynced], [<nil>] -> [True], [<nil>] -> [], [<nil>] -> []
  Normal   ApplicationConditionChanged  40m                 application-controller  [False] -> [True], [ChartVersionResolutionFailed] -> [], [failed to resolve chart version [name: "icanhazbroccoli-test", version: "0.0.2", repo: "http://0.0.0.0:8879"]: no chart version found] -> [Rolling out initial release "testapp-v1alpha1-821ad2b7-0"]
```


pkg/controller/application/application_controller.go
```diff
@@ -312,11 +312,17 @@ func (c *Controller) wrapUpApplicationConditions(app *shipper.Application, rels
...
 
 	abortingCond := apputil.NewApplicationCondition(shipper.ApplicationConditionTypeAborting, corev1.ConditionFalse, "", "")
-	apputil.SetApplicationCondition(&app.Status, *abortingCond)
+	if diff := apputil.SetApplicationCondition(&app.Status, *abortingCond); !diff.IsEmpty() {
+		c.reportApplicationConditionChange(app, diff)
+	}
 	validHistoryCond := apputil.NewApplicationCondition(shipper.ApplicationConditionTypeValidHistory, corev1.ConditionTrue, "", "")
-	apputil.SetApplicationCondition(&app.Status, *validHistoryCond)
+	if diff := apputil.SetApplicationCondition(&app.Status, *validHistoryCond); !diff.IsEmpty() {
+		c.reportApplicationConditionChange(app, diff)
```
juliogreff: This seems a bit repetitive. What do you think about having a set of diffs in the same struct instead? Then we could check it only once at the end, and apply the relevant operations.

juliogreff:
I like the general approach here, would love to see this for all objects ❤️ 

I'm not sure about the output though, I'm afraid it'd be confusing to read. Maybe grouping the transition of the entire condition, instead of having granular fields, could help:

```
[RollingOut True] -> [RollingOut False ChartVersionResolutionFailed "failed to resolve chart version [name: "icanhazbroccoli-test", version: "0.0.2", repo: "http://0.0.0.0:8879"]: no chart version found"]
```

Note: I'm not too sure about omitting empty fields like I did is helpful or not (it may not be, can be just as confusing).

icanhazbroccoli:
@juliogreff thanks a lot for your feedback. I will take your notes into account and continue working on it <3

icanhazbroccoli:
A new representation looks like:

```
Events:
  Type     Reason                       Age                 From                    Message
  ----     ------                       ----                ----                    -------
  Normal   ApplicationConditionChanged  39s                 application-controller  [] -> [RollingOut False "ChartVersionResolutionFailed" "failed to resolve chart version [name: \"icanhazbroccoli-test\", version: \"0.0.2\", repo: \
"http://0.0.0.0:8879\"]: no chart version found"]
  Warning  FailedApplication            21s (x12 over 38s)  application-controller  failed to resolve chart version [name: "icanhazbroccoli-test", version: "0.0.2", repo: "http://0.0.0.0:8879"]: no chart version found
  Normal   ApplicationConditionChanged  15s                 application-controller  [] -> [Blocked False]
  Normal   ApplicationConditionChanged  14s                 application-controller  [] -> [Aborting False], [] -> [ValidHistory True], [] -> [ReleaseSynced True], [RollingOut False "ChartVersionResolutionFailed" "failed to resolve chart v
ersion [name: \"icanhazbroccoli-test\", version: \"0.0.2\", repo: \"http://0.0.0.0:8879\"]: no chart version found"] -> [RollingOut True "Rolling out initial release \"testapp-v1alpha1-821ad2b7-0\""]
```

pkg/util/application/conditions.go
```diff
@@ -1,13 +1,70 @@
...
 	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	shipper "github.com/bookingcom/shipper/pkg/apis/shipper/v1alpha1"
-	"sort"
+	diffutil "github.com/bookingcom/shipper/pkg/util/diff"
 )
 
+type ConditionDiff struct {
+	cond1, cond2 *shipper.ApplicationCondition
+}
```
juliogreff: I wonder if it isn't time we introduce an interface for conditions? Or, less preferred because of type safety, consolidate everything in a `Condition` type? Besides type safety, I see very little value in having to duplicate this type of code for every object we have.

icanhazbroccoli: I assume, the early development of these conditions was inspired by the native k8s approach (e.g.: https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/types.go#L2253). I wouldn't really challenge this approach of k8s and preserve it the way it is in shipper as well. From this perspective an interface looks like a decent alternative.

icanhazbroccoli:
@juliogreff @parhamdoustdar @hbarkal hey folks, I'd really like to hear from you on the recent changes I made here. The approach is very conservative compared to some alternatives that I had in mind initially. Some spots look a bit clumsy, but that doesn't look like a showstopper right now. I'd be very delighted to hear your opinion before we agree this is the way to go and replicate the approach to other controllers.

pkg/util/condition/condition.go
```diff
@@ -0,0 +1,172 @@
...
+
+func condEqual(c1, c2 Condition) bool {
+	if condIsNil(c1) && condIsNil(c2) {
+		return true
+	}
+	if condIsNil(c1) || condIsNil(c2) {
+		return false
+	}
+	// Clumsy :-(
+	switch c1t := c1.(type) {
```
juliogreff: I don't mind this for right now. If this lives on, though, the failure mode is gonna be someone introducing a new type whose values will always be different. I think we should fail as loud as possible (a panic, perhaps?) when that happens, to save someone's debugging time.

juliogreff:
I'm not thrilled about the reflection bits, but I think this delivers enough value, and the interfaces are contained enough, that we can address this in a different PR if we really want to. I'll take a closer look at the nuts and bolts on Monday, but I think that at a high level you're on point here 👍

icanhazbroccoli:
Reverted the "WIP" status due to my discovery of pkg/conditions module: this seems to be a older alternative approach for condition modification. My short-term intent is to get rid of this package completely in order to achieve a unification among our modules.

pkg/controller/traffic/traffic_controller.go
```diff
@@ -266,48 +283,38 @@ func (c *Controller) syncHandler(key string) error {
...
-					err.Error())
+				reason = UnknownError
 			}
+			cond := trafficutil.NewClusterTrafficCondition(
+				shipper.ClusterConditionTypeReady,
+				corev1.ConditionFalse,
+				reason,
+				err.Error(),
+			)
+			diff.Append(trafficutil.SetClusterTrafficCondition(clusterStatus, *cond))
```
juliogreff: ❤️ ❤️ ❤️

pkg/util/capacity/conditions.go
```diff
@@ -0,0 +1,126 @@
...
+	for _, ch := range chunks {
+		if len(ch) > 0 {
+			if b.Len() > 0 {
+				b.WriteByte(' ')
+			}
+			b.WriteString(ch)
+		}
+	}
+	return b.String()
+}
```
juliogreff: I'd say that with funcs that mostly deal with primitives like this one and `String()`, you could extract the logic to an util func that these just delegate to, to cut down on duplication a little bit.

Issue #23, opened 2018-10-01
Check the production Service Endpoint object to see whether traffic has progressed


isutton:
The actual implementation considers that if a Pod has been successfully patched to be added into the Load Balancer, the Pod is receiving traffic.

This is not reliable, since the production Service `.spec.selector` is modified, we risk that existing Pods do not have matching labels and traffic will never reach those Pods, resulting in Shipper being stuck while reporting erroneously that traffic is being routed to those Pods.

PR #219, opened 2019-10-16
traffic controller: preparatory refactoring

juliogreff:
I was trying to start work on #23, but I was surprised by an assortment of small issues that got in the way. In order to keep PRs smaller and atomic, I'm opening this one before I continue on the actual work.


pkg/controller/eventhandler.go
```diff
@@ -1,13 +1,12 @@
...
 
 import (
-	shipper "github.com/bookingcom/shipper/pkg/apis/shipper/v1alpha1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog"
 )
 
-func NewAppClusterEventHandler(callback func(obj interface{})) cache.ResourceEventHandler {
+func NewAppClusterEventHandler(filterLabel string, callback func(obj interface{})) cache.ResourceEventHandler {
```
icanhazbroccoli: Shall we start thinking about introducing a more abstract Filter substance here that can carry multiple labels etc?

juliogreff: Yes and no. I think I got the abstraction wrong here in the first place. I pushed 35d7f2d, which is probably not what you meant, but I think it addresses the issue anyway. What do you think?

PR #220, opened 2019-10-16
Upgraded to helm 2.15.0 fixes #102

### LINKED ISSUE #102: Large integer `image tag` gets converted to scientific notation
When an `image tag` is a large integer (e.g. `20190612073634`) it gets converted to the scientific notation (e.g `2.0190612073634e+13`)

For example the following `release` configuration:
```yaml
Values:
  Image:
    Repository:  [...]
    Tag:         20190612073634
```

Results in this `pod` configuration:
```yaml
Containers:
  app:
    Container ID:
    Image:          [...]:2.0190612073634e+13
```

As a result, the pod is at `InvalidImageName` error state

icanhazbroccoli:
This commit initialises our transition to the updated version of Helm
(2.15+) which should solve our old issue with yaml converting digit-only
docker image tags to scientific notation and breaks it all around.

The commit that fixes it was merged into helm v2 in version 2.15.0:
https://github.com/helm/helm/commit/9014bd9c502d92fb78bfee553abdba5cf35e878f

The PR is going to stay in draft mode until helm 2.15 stable is released.


icanhazbroccoli:
I'm closing this PR because we had to revert the changes in Helm: both dev-v2 and dev-v3. The change caused a major regression in numeric functions (see https://github.com/helm/helm/issues/6708 for more details). Our attempt to close the breach turned into an unreasonably massive and ugly change https://github.com/helm/helm/pull/6709, which didn't look like a small fix at all. Taking into consideration potential risks, the fixing PR has been closed with no merge and the initial change reverted. There is a possibility the problem would be fixed in the future versions of Helm3 (the idea is to use reflect-based function overloads; we've made some initial progress here: https://github.com/icanhazbroccoli/helm/commit/8b359af3c3e246e90c0e3f7667be494f19d8e716, but it's not close to anything working yet). My apologies for being over-optimistic and wasting so much time on it.

PR #226, opened 2019-10-28
Webhook: validate block override anotation

hbarkal:
Webhook should validate existence of rollout block when added to override annotation of an application or a release

### DIFF
--- a/pkg/util/rolloutblock/blocks.go
+++ b/pkg/util/rolloutblock/blocks.go
@@ -1,8 +1,6 @@
 package rolloutblock
 
 import (
-	"regexp"
-
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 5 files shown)


pkg/util/rolloutblock/blocks.go
```diff
@@ -63,18 +61,34 @@ func BlocksRollout(rolloutBlockLister shipperlisters.RolloutBlockLister, obj met
 	}
 }
 
-func ValidateOverrides(overrides ObjectNameList) error {
-	if len(overrides) == 0 {
+func ProcessBlocks(existingBlocks ObjectNameList, overrides ObjectNameList) error {
```
icanhazbroccoli: I suggest to use more consistent naming for arguments of the same time: e.g.:
```go
func ProcessBlocks(existing, overrides ObjectNameList)
```

The method name needs a comment: the normal workflow here is to return an error, and the meaning of the nil return result is not trivial. I believe we might leave a comment explaining what return values mean.

The method name doesn't really process blocks: it figures out if it's good to go or not. From this perspective we might consider renaming it to something more self-explanatory.

pkg/util/rolloutblock/blocks.go
```diff
@@ -63,18 +61,34 @@ func BlocksRollout(rolloutBlockLister shipperlisters.RolloutBlockLister, obj met
...
+	effectiveBlocks := existingBlocks.Diff(overrides)
+	if len(effectiveBlocks) == 0 {
 		return nil
+	} else {
```
icanhazbroccoli: Suggested edit: drop the else statement as the nil-case return returns instantly.

PR #227, opened 2019-11-05
Release object status gained a fleet summary report fixes #139

### LINKED ISSUE #139: Improve Reporting In the Release Object About the Fleet
Now that we have the fleet summary report in the CapacityTarget object, we can summarise it in a way that is easy to read in the conditions of the Release object.

We need to figure out the answers to the following questions:

- How can we use the fleet summary to figure out what's wrong?
- Once we figure it out, how can we turn it into a string short enough to be put into the `message` of a condition?

icanhazbroccoli:
This commit introduces a new change to Shipper API: the release object
status gained a fleet summary report. The source of the data is mainly
coming from a relevant capacity target object fleet summary.

An example of a happy report:
```
  reports:
  - achievedPercent: 100
    availableReplicas: 2
    cluster: docker-desktop
    summary: |-
      2 container(s) [nginx(x2)] are in status: ContainersReady:True
      2 container(s) [nginx(x2)] are in status: Initialized:True
      2 container(s) [nginx(x2)] are in status: PodScheduled:True
      2 container(s) [nginx(x2)] are in status: Ready:True
```

An example of an unhappy report:
```
  reports:
  - cluster: docker-desktop
    problems: 'PodScheduled:False, reason: Unschedulable, message: 0/1 nodes are available:
      1 Insufficient cpu, 1 Insufficient memory.'
```


pkg/util/release/reports.go
```diff
@@ -0,0 +1,63 @@
...
+				status))
+		}
+	}
+
+	sort.Strings(res)
+
+	return strings.Join(res, "\n")
+}
+
+func BuildClusterProblemReport(s []shipper.PodStatus) string {
```
juliogreff: I worry that problems at a pod level don't tell the entire story. Problems with quotas or a validating webhook denying the creation of pods, for instance, won't show up. I suppose we'll need further changes to the capacity controller to surface that up (presumably by looking at Deployment conditions) before doing changes here though.

icanhazbroccoli: Thanks for sharing this @juliogreff! I wonder if you see this reasonable to make the release controller to have an application cluster connection pool and communicate to them directly or would you suggest an alternative approach to get deployment condition insights?

pkg/apis/shipper/v1alpha1/types.go
```diff
@@ -235,6 +236,14 @@ type ReleaseCondition struct {
 	Message            string                 `json:"message,omitempty"`
 }
 
+type ReleaseClusterReport struct {
```
juliogreff: This PR is only concerned with generating reports from CapacityTargets, and that's ok. We do need to think about how the data structure is going to evolve when we start generating reports for InstallationTargets and TrafficTargets.

Issue #71, opened 2019-03-07
Release Controller Should Enforce State For All Releases Based On Their Successor If Found


parhamdoustdar:
Right now Shipper only reacts to the last two releases, the incumbent and contender. This means that if there is a problem in a release that is not either the incumbent and contender, it will stay there for ever.

This issue showed itself with a Release object that was supposed to be completely rolled out. however, for some reason, the `spec` of the CapacityTarget and TrafficTarget were still saying that this release should have _10%_ capacity and _10%_ traffic.

In a perfect world, Shipper would do the following:

1. Receive an update event for an old Release
2. Find out the next release in the history, the _newer_ Release
3. Figure out what state the associated target objects (`InstallationTarget`, `CapacityTarget`, and `TrafficTarget`) of the _old_ release should be, based on the _newer_ release
4. Apply that to the target objects

icanhazbroccoli:
This issue is planned to be fixed with the upcoming merge request #70. The relevant e2e tests are to be delivered asap.

icanhazbroccoli:
Implemented in https://github.com/bookingcom/shipper/pull/254. Will close the ticket once it gets merged (due release version: 0.8)

icanhazbroccoli:
#254 got merged into master, closing this ticket.

PR #235, opened 2019-11-26
ReleaseController scheduler only keeps 2 recent target pairs

icanhazbroccoli:
This commit introduces a change in release controller scheduler. It
proposes to only keep 2 recent target pairs (e.g. 2 latest capacity
targets, 2 latest traffic targets and 2 latest installation targets).
This is mainly done as a step towards simplifying Shipper's job on
handling accidental wakeups of old releases (see
https://github.com/bookingcom/shipper/issues/71 for more details).

### DIFF
--- a/pkg/controller/capacity/capacity_controller.go
+++ b/pkg/controller/capacity/capacity_controller.go
@@ -2,12 +2,12 @@ package capacity
 
 import (
 	"fmt"
-	"reflect"
 	"sort"
 	"time"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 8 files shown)


kanatohodets:
Something to ponder: deleting the Target objects means nuking the InstallationTarget, which will cause cleanup of the chart objects in the application clusters.

This might be too much trouble, but IMO ideally we figure out how to support the history of Target objects so that we can work towards reducing the amount of logic which gives special treatment to 'contender' and 'incumbent'. This was what we were trying to describe in #71: we would like to devise a general rule for "successor/predecessor" that applies to all Releases, rather than only reconciling the incumbent and contender.

An example use case that I would like to be able to ultimately support: something like https://brandon.dimcheff.com/2018/02/rainbow-deploys-with-kubernetes/, where you might turn down the traffic for the old release, but you keep the pods in place in order to give them time to transfer state / keep serving connected clients. This means you might have "the incumbent's incumbent" with > 0 pods, so cleaning up the InstallationTarget would be destructive.

YAGNI is a totally fair response to this feedback, but IMO a healthy direction for Shipper's code is to gradually remove or 'contain' special casing for incumbent / contender, so this change spooks me slightly.

PR #238, opened 2019-11-28
traffic controller: check progression through Endpoints

### LINKED ISSUE #23: Check the production Service Endpoint object to see whether traffic has progressed

juliogreff:
Our current traffic controller considers a Pod being successfully
patched to have traffic labels as traffic actually progressing.

This isn't reliable for several reasons:

* A Service's `.spec.selector` can be modified, resulting in Pods no
longer receiving traffic, but the traffic controller will still report
that traffic has been achieved.

* Pods can be in a non-Ready state, making them not receive traffic, and
again Shipper will report traffic as being achieved.

To solve both of those issues, we now watch endpoints and extract actual
traffic information from them. For our purposes, only Ready Pods will
count towards traffic. This might make rollouts slightly slower (as
traffic now has to wait for pods to be actually ready to receive
traffic) but also much more reliable: we won't drain traffic from an
incumbent Release until the contender is actually fully ready.

Closes #23.

### DIFF
--- a/pkg/controller/traffic/pod_label_shifter.go
+++ b/pkg/controller/traffic/pod_label_shifter.go
@@ -18,8 +18,8 @@ import (
 
 type podLabelShifter struct {
 	appName               string
+	releaseName           string
 	namespace             string
-	serviceSelector       labels.Selector
 	clusterReleaseWeights clusterReleaseWeights
 }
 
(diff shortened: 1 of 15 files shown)


pkg/controller/traffic/pod_label_shifter.go
```diff
@@ -18,159 +18,168 @@ import (
...
 	appPods, err := podLister.List(appSelector)
 	if err != nil {
 		return 0, shippererrors.NewKubeclientListError(
 			podGVK, p.namespace, appSelector, err)
 	}
 
-	releasePods, err := podLister.List(releaseSelector)
+	releaseSelector := labels.Set{shipper.ReleaseLabel: p.releaseName}.AsSelector()
+
+	svc, err := p.getService(informerFactory)
```
icanhazbroccoli: This implies on a singular active service per release, which is very true now. Might be worth leaving a comment this assumption was kept in mind. If this constraint would be challenged in the future, we should remember about this spot.

pkg/controller/traffic/pod_label_shifter.go
```diff
@@ -18,159 +18,168 @@ import (
...
-			podGVK, p.namespace, releaseSelector, err)
+		return 0, err
+	} else if svc.Spec.Selector == nil {
+		return 0, shippererrors.NewTargetClusterServiceMissesSelectorError(p.namespace, svc.Name)
 	}
 
-	var totalWeight uint32 = 0
-	for _, weight := range releaseWeights {
-		totalWeight += weight
+	trafficSelector := labels.Set(svc.Spec.Selector).AsSelector()
```
icanhazbroccoli: (the question is most likely powered by my ignorance) Can svc.Spec.Selector be used without wrapping it in a set and selector? Conceptually, all looks good, it's just the statement that looks a bit confusing.

pkg/controller/traffic/pod_label_shifter.go
```diff
@@ -18,159 +18,168 @@ import (
...
-	var trafficPods []*corev1.Pod
-	var idlePods []*corev1.Pod
-	for _, pod := range releasePods {
-		if getsTraffic(pod, trafficSelector) {
-			trafficPods = append(trafficPods, pod)
-			continue
+		status.ReleasePods, releaseTargetWeight, len(appPods), totalTargetWeight)
+
+	// TODO(jgreff): we need to be able to tell, from the outside, that all
+	// pods have been labeled, but some may not be ready yet.
```
icanhazbroccoli: Do you mean some kind of annotations?

pkg/controller/traffic/pod_label_shifter.go
```diff
@@ -18,159 +18,168 @@ import (
...
 	patches := make(map[string][]byte)
 
-	if len(trafficPods) > targetPods {
-		excess := len(trafficPods) - targetPods
+	if len(status.LabeledPods) > targetPods {
+		excess := len(status.LabeledPods) - targetPods
 
 		for i := 0; i < excess; i++ {
-			pod := trafficPods[i].DeepCopy()
+			pod := status.LabeledPods[i].DeepCopy()
```
icanhazbroccoli: Are labeled pods deterministically sorted?

pkg/controller/traffic/pod_label_shifter.go
```diff
@@ -18,159 +18,168 @@ import (
...
+	services, err := informerFactory.Core().V1().Services().Lister().
+		Services(p.namespace).List(serviceSelector)
+	if err != nil {
+		return nil, shippererrors.NewKubeclientListError(
+			gvk, p.namespace, serviceSelector, err)
 	}
-	return true
+
+	if n := len(services); n != 1 {
+		return nil, shippererrors.NewUnexpectedObjectCountFromSelectorError(
```
icanhazbroccoli: It looks pretty explicit, awesome, thanks!

pkg/controller/traffic/traffic_shifting_status.go
```diff
@@ -0,0 +1,95 @@
...
+// TODO(jgreff): maybe we don't need the pods themselves in Ready/NotReady
+
+type TrafficShiftingStatus struct {
+	LabeledPods   []*corev1.Pod
+	UnlabeledPods []*corev1.Pod
+
+	ReadyPods    int
+	NotReadyPods int
+
+	ReleasePods int
```
icanhazbroccoli: I wonder if naming should reflect that LabeledPods and ReadyPods are of different type (the attributes are public)

pkg/controller/traffic/utils_test.go
```diff
@@ -0,0 +1,104 @@
...
+// NOTE: this does not try to implement endpoint subsets at all. All
+// pods are assumed to be part of the same subset and to be always ready.
+func shiftPodInEndpoints(pod *corev1.Pod, endpoints *corev1.Endpoints) *corev1.Endpoints {
+	var podGetsTraffic bool
+	trafficStatusLabel, ok := pod.Labels[shipper.PodTrafficStatusLabel]
+	if ok {
+		podGetsTraffic = trafficStatusLabel == shipper.Enabled
+	}
+
+	addresses := endpoints.Subsets[0].Addresses
```
icanhazbroccoli: Are addresses sorted? If yes, a suggested edit is to use `sort.Search()`. If no, please, never mind :-)

pkg/controller/traffic/utils_test.go
```diff
@@ -0,0 +1,104 @@
...
+			TargetRef: &corev1.ObjectReference{
+				Kind:      "Pod",
+				Namespace: pod.Namespace,
+				Name:      pod.Name,
+			},
+		}
+
+		endpoints.Subsets[0] = corev1.EndpointSubset{
+			Addresses: append(addresses, address),
+		}
```
icanhazbroccoli: Are we guaranteed to never loose any additional attributes from `endpoints.Subsets[0]`? What i mean is: this structure is being re-created with the only attribute `Addresses`. Any chance it would be existing with some other data that might be accidentally lost?

icanhazbroccoli:
@juliogreff great job, man! I left this PR unmerged as I had a few non-critical comments. Please have a look once you have a chance and merge it if none of them are essential.

pkg/controller/traffic/pod_label_shifter.go
juliogreff: `svc.Spec.Selector` is a `map[string]string`. k8s types are a mess :'( Not even in other types `Spec.Selector` is a `map[string]string` even.

pkg/controller/traffic/pod_label_shifter.go
juliogreff: No, I mean what was achieved in https://github.com/bookingcom/shipper/pull/245/commits/a9e5162b0ac975863e9a36b7062c526bb3685d27 -- basically, returning the trafficStatus object so the controller can use it to report things better.

pkg/controller/traffic/traffic_shifting_status.go
juliogreff: I agree. These got changed in later PRs though, si hopefully that addresses your concerns better than here

pkg/controller/traffic/utils_test.go
juliogreff: Since this is a test environment, we control the object entirely. So I guess not, not guaranteed to be sorted.

PR #239, opened 2019-12-02
capacity controller: ensure Operational condition is always set

juliogreff:
In a few cases, capacity controller wouldn't set the Operational
condition in CapacityTargets, even though it could operate on them
successfully, and even report non-Ready states!


juliogreff:
Further changes to this were needed and will be applied in #242 instead, so I'm closing this one.

PR #242, opened 2019-12-03
Create a `Ready` condition in CapacityTarget to inform Release Controller when capacity has been achieved

juliogreff:
This builds on top of work done in #245 which builds on top of #241, which builds on top of #238, so make sure those are merged before going through the changes here :)

We're moving towards relocating the smarts in the checks for a target's
readiness away from the executor and into the controllers themselves,
leaving the executor to check for a single Ready condition at the
target's top level status. This PR achieves exactly that for
CapacityTarget objects.


pkg/controller/capacity/capacity_controller.go
```diff
@@ -168,116 +172,122 @@ func (c *Controller) processCapacityTargetOnCluster(
...
-			"",
-			"")
-		diff.Append(capacityutil.SetClusterCapacityCondition(status, *condOperational))
-	}
+	readyCond = capacityutil.NewClusterCapacityCondition(
+		shipper.ClusterConditionTypeReady,
+		corev1.ConditionFalse,
+		PodsNotReady,
+		msg,
+	)
```
icanhazbroccoli: Not sure I red this correctly: no matter if there are some or no sad pods, ClusterConditionTypeReady is always being set to false here, right?

juliogreff: Correct. This happens, for instance, when you exceed quotas: you can't create pods, so they won't be sad, but you still don't have enough pods.

pkg/testing/controller_test_fixture.go
```diff
@@ -0,0 +1,63 @@
...
+	"fmt"
+
+	kubefake "k8s.io/client-go/kubernetes/fake"
+	"k8s.io/client-go/tools/record"
+
+	shipperfake "github.com/bookingcom/shipper/pkg/client/clientset/versioned/fake"
+	shipperinformers "github.com/bookingcom/shipper/pkg/client/informers/externalversions"
+)
+
+type ControllerTestFixture struct {
```
icanhazbroccoli: <3 this

pkg/controller/capacity/capacity_controller.go
```diff
@@ -168,116 +172,122 @@ func (c *Controller) processCapacityTargetOnCluster(
...
+	reports = []shipper.ClusterCapacityReport{*report}
 
-	if targetDeployment.Spec.Replicas == nil || rCnt != *targetDeployment.Spec.Replicas {
-		_, err = c.patchDeploymentWithReplicaCount(ct, targetDeployment, spec.Name, rCnt, status)
+	desiredReplicas := int32(replicas.CalculateDesiredReplicaCount(uint(spec.TotalReplicaCount), float64(spec.Percent)))
+	if deployment.Spec.Replicas == nil || desiredReplicas != *deployment.Spec.Replicas {
+		_, err = c.patchDeploymentWithReplicaCount(deployment, spec.Name, desiredReplicas)
 		if err != nil {
+			readyCond = capacityutil.NewClusterCapacityCondition(
+				shipper.ClusterConditionTypeOperational,
```
hihilla: This might be a mistake, readyCond  = ... ClusterConditionType*Operational* ... ?

PR #243, opened 2019-12-04
Generate OpenAPI Specs

zoidbergwill:
Some of the Kubernetes clients use clients generated from a swagger.json.

kubernetes-client/javascript uses kubernetes-client/gen which uses
OpenApiTools/openapi-generator and kubernetes/kubernetes/api/openapi-spec/swagger.json

There are open/closed/rotten/ignored issues all over the place about
making these easier to use with CRDs.

Wrongly closed issue on kubernetes-client/javascript: kubernetes-client/javascript#341
Open issue on kubernetes-client/gen: kubernetes-client/gen#121

It seems like kubernetes/kube-openapi is one of the repos working on
solving the main part of this problem, generating the Go code to make
it easy to generate a swagger.json

I found the following Argo code via the following issue on kube-openapi: https://github.com/kubernetes/kube-openapi/issues/13

https://github.com/argoproj/argo/blob/master/hack/gen-openapi-spec/main.go
https://github.com/argoproj/argo/blob/master/pkg/apis/workflow/v1alpha1/openapi_generated.go#L14:6
https://github.com/kubernetes/kube-openapi/blob/master/cmd/openapi-gen/openapi-gen.go

So this PR adds similar stuff to Shipper.

gen-openapi-spec uses the generated openapi_generated.go module that is
generated via the openapi-gen command, to generate a `swagger.json`

The following commits:
- add the code to generate openapi_generated.go
- add the code to generate swagger.json from openapi_generated.go


zoidbergwill:
So with the related **swagger.json**, in kubernetes-client/gen I can do the following:

Create a file called shipper_settings.sh in the openapi directory with the following contents:

```
export KUBERNETES_BRANCH="master"
export CLIENT_VERSION="1.0.0b1"
export PACKAGE_NAME="shipper-client"
export USERNAME="zoidbergwill"
export REPOSITORY="shipper"
```

Then generate a Typescript client using the following

```
./typescript.sh shipper-client shipper_settings.sh  
```

And I should be able to use it with kubernetes-client/javascript, by doing something like:

```javascript
const k8s = require('@kubernetes/client-node');
const shipper = require('@booking/shipper-client');

const kc = new k8s.KubeConfig();
kc.loadFromDefault();

const k8sApi = kc.makeApiClient(shipper.V1alpha1Api);
```

zoidbergwill:
So in my last commit I manually added some other k8s.io types that we use. I don't know how to get them into our generated openAPI code. I'll investigate what argo is doing.

https:/github.com/bookingcom/shipper/blob/master/pkg/apis/shipper/v1alpha1/types.go?expand=1#L6-L7

PR #245, opened 2019-12-09
traffic controller: report the correct achieved traffic weight

juliogreff:
This builds on top of work done in #241 , which builds on top of #238, so make sure those are merged before going through the changes here :)

Before this, the pod label shifter would return a bogus achieved traffic
weight whenever the number of pods required to take traffic was equal to
the number of pods ready. Fixing that up front would've broken the
release controller, as it used to check the achieved traffic weight
instead of just looking at conditions (which at the time didn't exist).
Now that this has been fixed [1], we can start reporting the actual
achieved weight instead of having to lie.

The diff is also a fair bit larger than it should be, since I got a bit
carried away with the pod label shifter refactoring. There's a
rationale, though: it moves all the actual logic of traffic calculations
away from ever touching informers and clients and into
traffic_shifting_status.go, making unit testing much easier.
pod_label_shifter.go now only does the mechanical job of shifting
labels, making its own surface area much smaller. As a bonus, tests are
much faster too!

Last but not least, now we can also tell the difference traffic shifting
being in progress (as in, shipper is currently moving labels around) and
it being stuck (as in, pods aren't ready). This was also some extra
motivation for the state we accumulate in trafficShiftingStatus.

[1] https://github.com/bookingcom/shipper/pull/241

### DIFF
--- a/pkg/controller/traffic/traffic_controller.go
+++ b/pkg/controller/traffic/traffic_controller.go
@@ -8,6 +8,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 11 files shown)


pkg/controller/traffic/traffic_controller.go
```diff
@@ -311,97 +311,160 @@ func (c *Controller) processTrafficTargetOnCluster(
...
+		if err != nil {
+			readyCond = trafficutil.NewClusterTrafficCondition(
+				shipper.ClusterConditionTypeReady,
+				corev1.ConditionFalse,
+				InternalError,
+				err.Error(),
+			)
+
+			return err
+		} else {
```
icanhazbroccoli: super minor remark: else could be omitted

pkg/controller/traffic/traffic_controller.go
```diff
@@ -311,97 +311,160 @@ func (c *Controller) processTrafficTargetOnCluster(
...
+			)
+		}
 	} else {
 		readyCond = trafficutil.NewClusterTrafficCondition(
 			shipper.ClusterConditionTypeReady,
 			corev1.ConditionFalse,
 			PodsNotReady,
-			fmt.Sprintf("weight expected: %d, weight achieved: %d", spec.Weight, achievedWeight),
+			fmt.Sprintf(
+				"%d out of %d pods are not ready. this might require intervention, check the CapacityTarget for more information",
```
icanhazbroccoli: We might consider using this guide we propose in release controller: "try using kubectl get ct ..." for the sake of consistency here.

PR #250, opened 2019-12-19
Report installation progression through a Ready condition

juliogreff:
This builds on top of work done in #242, which builds on top of #245, which builds on top of #241, which builds on top of #238, so make sure those are merged before going through the changes here :)

We're moving towards relocating the smarts in the checks for a target's
readiness away from the executor and into the controllers themselves,
leaving the executor to check for a single `Ready` condition at the
target's top level status. This commit does exactly that for
InstallationTarget objects, which should be the last step in this
process.


pkg/controller/release/strategy_executor.go
```diff
@@ -67,26 +67,16 @@ func genInstallationEnforcer(curr, succ *releaseInfo) PipelineStep {
...
+		if ready, clusters := checkInstallation(curr.installationTarget); !ready {
+			cond.SetFalse(
+				shipper.StrategyConditionContenderAchievedInstallation,
+				conditions.StrategyConditionsUpdate{
+					Reason:             ClustersNotReady,
+					Message:            fmt.Sprintf("clusters pending installation: %v. for more details try `kubectl describe it %s`", clusters, curr.installationTarget.Name),
+					Step:               targetStep,
+					LastTransitionTime: time.Now(),
+				},
+			)
```
icanhazbroccoli: <3

pkg/controller/installation/installation_controller.go
```diff
@@ -282,166 +295,155 @@ func (c *Controller) getInstallationTargetForReleaseAndNamespace(release, namesp
...
-			if s.Name == name {
-				existingConditionsPerCluster[name] = s.Conditions
-			}
-		}
-	}
-	return existingConditionsPerCluster
+	return nil
 }
 
 func (c *Controller) GetClusterAndConfig(clusterName string) (kubernetes.Interface, *rest.Config, error) {
```
hihilla: I think this was meant to be called GetClientAndConfig?

Issue #125, opened 2019-07-24
Scheduler does not update "Scheduled" condition in all cases


juliogreff:
We had an interesting outage where our charts server went down during a Shipper upgrade, so Shipper could no longer fetch and render charts either from the cache or from the server. This changed all releases to `Scheduled: false` in the next resync, as was expected. When the charts server recovered, though, Shipper *did not* update the `Scheduled` condition back to `true`, even though the releases were indeed scheduled properly.

PR #254, opened 2020-01-06
Release controller enforces release state based on their successor

icanhazbroccoli:
This commit introduces some major changes that relax our paradigm or
incumbent-contender release relatoinships.

First of all, release controller lost application sub-controller. The
latter was serving as an enclosing component in release reconciliation
loop ensuring 2 head releases are alwais in a consistent state
(incumbent and contender). The previous version of this code was dealing
with these 2 releases no matter what release initiated the
reconciliation loop. This approach was changed for the sake of ensuring
the state of historical releases. We had this situation in the past
where a historical release was activated due to a human mistake or a CI
glitch which was causing some unwanted system behavior and shipper was
unable to get it fixed.

Having that said, all releases are now treated equally. The major
constraint release controller uses is predcessor-ancestor relationship
between releases. Predcessor state is defined by its ancestor's state.
If none exists, it is defined by the state spec.

Release reconciliation chain became a bit more agressive: every change
on a release object will trigger a reconciliation loop for ancestor
and predcessor releases. Therefore we might notice some more noise in
this chain in production.

Some components still heavily depend on contender-incumbent constraint,
therefore left as is for the sake of making this change fairly compact
and traceable. It must be a subject for future refactorings.


pkg/controller/release/application.go
```diff
@@ -1,223 +0,0 @@
...
-			strategy:  strategy,
-		}, nil
-	}
-
-	incumbentReleaseInfo, err := c.buildReleaseInfo(incumbentRelease)
-	if err != nil {
-		return nil, err
-	}
-
-	return &Executor{
```
juliogreff: Unless I missed it, looks like the old `Executor` is not used anymore. Should we delete it as well?

pkg/util/release/generation.go
```diff
@@ -27,3 +28,54 @@ func SetGeneration(release *shipper.Release, generation int) {
...
+	var releaseIx = len(appReleases)
+	for ix := range appReleases {
+		if release.Name == appReleases[ix].Name &&
+			release.Namespace == appReleases[ix].Namespace {
+			releaseIx = ix
+			break
+		}
+	}
+
+	if releaseIx < len(appReleases) { // release is in the list
```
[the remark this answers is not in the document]
icanhazbroccoli: Fair question! It happens when a release was deleted and we still want to trigger a reconciliation loop in the right location (the very first implementation re-queued all releases)

pkg/controller/release/release_controller.go
```diff
@@ -283,30 +275,140 @@ func (c *Controller) syncOneReleaseHandler(key string) error {
...
+
+	_, err = c.ensureReleaseState(rel.DeepCopy())
+	if err != nil {
+		return err
+	}
 
-	if !reflect.DeepEqual(initialRel, rel) {
-		if _, err := c.clientset.ShipperV1alpha1().Releases(namespace).Update(rel); err != nil {
-			return shippererrors.NewKubeclientUpdateError(rel, err).
-				WithShipperKind("Release")
```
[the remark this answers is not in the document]
icanhazbroccoli: Very good point. I'll try to improve on it. Thanks!

pkg/controller/release/release_controller.go
```diff
@@ -67,8 +72,7 @@ type Controller struct {
 	rolloutBlockLister shipperlisters.RolloutBlockLister
 	rolloutBlockSynced cache.InformerSynced
 
-	releaseWorkqueue     workqueue.RateLimitingInterface
-	applicationWorkqueue workqueue.RateLimitingInterface
```
juliogreff: There's a comment a few lines above that says "Release Controller has 2 primary workqueues: releases and applications" - I guess that can go away now ^_^

pkg/controller/release/strategy_executor.go
```diff
@@ -0,0 +1,470 @@
...
+	return func(e *StrategyExecutor, cond conditions.StrategyConditionsMap) (PipelineContinuation, []ExecutorResult, []ReleaseStrategyStateTransition, error) {
+		strategy := curr.release.Spec.Environment.Strategy
+		targetStep := curr.release.Spec.TargetStep
+		isLastStep := int(targetStep) == len(strategy.Steps)-1
+
+		if targetStep >= int32(len(strategy.Steps)) {
+			err := fmt.Errorf("no step %d in strategy for Release %q",
+				targetStep, controller.MetaKey(curr.release))
+			return PipelineBreak, nil, nil, shippererrors.NewUnrecoverableError(err)
+		}
```
juliogreff: It seems like this is the only error that can happen inside the enforcers, and it's the exact same thing across all of them. Would it be possible to extract choosing the strategy step to `Execute`, and not even generate a pipeline if there's an error there? It sounds to me that it would simplify things a little bit.

pkg/controller/release/release_controller.go
```diff
@@ -283,30 +275,146 @@ func (c *Controller) syncOneReleaseHandler(key string) error {
...
+		diff := apputil.SetApplicationCondition(&app.Status, *releaseSyncedCond)
+		_, err = c.clientset.ShipperV1alpha1().Applications(app.Namespace).Update(app)
+		if err != nil {
+			return nil, shippererrors.NewKubeclientUpdateError(app, err).
+				WithShipperKind("Application")
+		}
+		if !diff.IsEmpty() {
+			c.recorder.Eventf(app, corev1.EventTypeNormal, "ApplicationConditionChanged", diff.String())
+		}
+		return nil, err
```
juliogreff: This doesn't actually work: it'll trigger a new run of the application controller, which will set this condition back to True, since it doesn't know much about strategies and will assume everything is alright.

I think a better approach would be to update the release instead, if it's the head, which is where an user would expect to see strategy errors when rolling out.

pkg/controller/release/release_controller.go
```diff
@@ -283,30 +275,146 @@ func (c *Controller) syncOneReleaseHandler(key string) error {
...
+	}
+	if succ != nil {
+		relInfoSucc, err = c.buildReleaseInfo(succ)
+		if err != nil {
+			return nil, err
 		}
 	}
 
+	//TODO: this flag is deprecated, we keep it here for the sake of gradual
+	//changes. The flag propagates as deep as
```
juliogreff: as deep as ...? :)

pkg/controller/release/strategy_executor.go
```diff
@@ -0,0 +1,470 @@
...
+	pipeline.Enqueue(genCapacityEnforcer(e.curr, e.succ))
+	pipeline.Enqueue(genTrafficEnforcer(e.curr, e.succ))
+
+	if isHead {
+		if hasTail {
+			pipeline.Enqueue(genTrafficEnforcer(e.prev, e.curr))
+			pipeline.Enqueue(genCapacityEnforcer(e.prev, e.curr))
+		}
+		pipeline.Enqueue(genReleaseStrategyStateEnforcer(e.curr, nil))
+	}
```
juliogreff: I have nothing to add, but I just love this ❤️Both the pipeline itself and the enforcer generators. This solves a major gripe I had with the previous executor, which made refactoring each step a pain in the ass.

pkg/controller/release/strategy_executor.go
```diff
@@ -0,0 +1,470 @@
...
+
+		if ready, clusters := checkInstallation(curr); !ready {
+			if len(curr.installationTarget.Spec.Clusters) != len(curr.installationTarget.Status.Clusters) {
+				cond.SetUnknown(
+					shipper.StrategyConditionContenderAchievedInstallation,
+					conditions.StrategyConditionsUpdate{
+						Step:               targetStep,
+						LastTransitionTime: time.Now(),
+					},
+				)
```
juliogreff: After #250, this whole branch becomes obsolete, as the executor shouldn't know about internal details of the target objects.

pkg/controller/release/strategy_executor.go
```diff
@@ -0,0 +1,470 @@
...
+	releaseutil "github.com/bookingcom/shipper/pkg/util/release"
+)
+
+type StrategyExecutor struct {
+	curr, prev, succ *releaseInfo
+	recorder         record.EventRecorder
+	hasIncumbent     bool
+}
+
+func NewStrategyExecutor(curr, prev, succ *releaseInfo, recorder record.EventRecorder, hasIncumbent bool) (*StrategyExecutor, error) {
```
juliogreff: Do we really need to return an `error` anymore?

pkg/controller/release/strategy_executor.go
```diff
@@ -0,0 +1,470 @@
...
+		if targetStep >= int32(len(strategy.Steps)) {
+			err := fmt.Errorf("no step %d in strategy for Release %q",
+				targetStep, controller.MetaKey(curr.release))
+			return PipelineBreak, nil, nil, shippererrors.NewUnrecoverableError(err)
+		}
+
+		isLastStep := int(targetStep) == len(strategy.Steps)-1
+
+		if achieved, newSpec, clustersNotReady := checkCapacity(curr.capacityTarget, uint(capacityWeight)); !achieved {
+			e.info("release %q hasn't achieved capacity yet", curr.release.Name)
```
juliogreff: `info` already prints out the name of the current release by default, so there's no need to say `release %q has blah...` here and a few other places anymore, as that was only necessary when we were operating on two releases at once.

pkg/controller/release/release_controller.go
```diff
@@ -283,30 +275,146 @@ func (c *Controller) syncOneReleaseHandler(key string) error {
...
+		)
+	}
 
-	klog.V(4).Infof("Done processing Release %q", key)
+	if len(patches) == 0 {
+		klog.V(4).Infof("Strategy verified, nothing to patch")
+		return curr, nil
+	}
+
+	klog.V(4).Infof("Strategy has been executed, applying patches")
```
juliogreff: We should say for which release a strategy has been executed, otherwise we have no context. This might be a little bit harder and maybe not strictly necessary, but it would also be useful if we could know which kinds of patches we're about to apply.

pkg/controller/release/release_controller.go
```diff
@@ -283,30 +275,146 @@ func (c *Controller) syncOneReleaseHandler(key string) error {
...
+	for _, patch := range patches {
+		c.applyPatch(curr.Namespace, patch)
```
juliogreff: `applyPatch` can return errors, and we're not checking them here, which will prevent any type of retries.

pkg/controller/release/release_controller.go
```diff
@@ -283,34 +272,145 @@ func (c *Controller) syncOneReleaseHandler(key string) error {
...
+	//changes. The flag propagates as deep as strategy conditions
+	//consolidation functions
+	// see pkg/util/conditions/strategy.go for more details.
+	hasIncumbent := len(releases) > 1
+	executor := NewStrategyExecutor(relInfoCurr, relInfoPrev, relInfoSucc, c.recorder, hasIncumbent)
+
+	patches, trans, err := executor.Execute()
+	if err != nil {
+		releaseSyncedCond := releaseutil.NewReleaseCondition(
+			shipper.ReleaseConditionTypeScheduled,
```
juliogreff: Maybe `Scheduled` isn't the right condition here. When this error triggers, shipper will enter into a loop as the scheduler will set this back to `True`, this will change it to `False`, trigger a new sync, repeat.

pkg/controller/release/strategy_executor.go
icanhazbroccoli: Awesome observation. Thank you so much!

pkg/controller/release/release_controller.go
```diff
@@ -283,30 +272,156 @@ func (c *Controller) syncOneReleaseHandler(key string) error {
...
+			shipper.ReleaseConditionTypeStrategyExecuted,
+			corev1.ConditionFalse,
+			conditions.StrategyExecutionFailed,
+			fmt.Sprintf("failed to execute strategy: %q", err),
+		)
+		diff.Append(releaseutil.SetReleaseCondition(&rel.Status, *releaseStrategyExecutedCond))
+		return rel, nil, nil, err
+	}
+
+	if !releaseutil.ReleaseStrategyExecuted(rel) {
```
juliogreff: I'd rather set the condition unconditionally. Having a conditional around has proven to attract problems in the past (see #125 for an example), and also goes against our philosophy of not reading conditions that we own (pretty much for the same reasons).

pkg/controller/release/release_controller.go
```diff
@@ -368,19 +496,21 @@ func (c *Controller) scheduleRelease(rel *shipper.Release) (*shipper.Release, er
...
 
 	rel = scheduledRel
-	condition = releaseutil.NewReleaseCondition(
-		shipper.ReleaseConditionTypeScheduled,
-		corev1.ConditionTrue,
-		"",
-		"",
-	)
-	diff.Append(releaseutil.SetReleaseCondition(&rel.Status, *condition))
+	if !releaseutil.ReleaseScheduled(rel) {
```
juliogreff: Same as my previous comment on setting conditions unconditionally.

pkg/controller/release/release_controller.go
```diff
@@ -283,30 +272,156 @@ func (c *Controller) syncOneReleaseHandler(key string) error {
...
+	if err != nil {
+		releaseStrategyExecutedCond := releaseutil.NewReleaseCondition(
+			shipper.ReleaseConditionTypeStrategyExecuted,
+			corev1.ConditionFalse,
+			conditions.StrategyExecutionFailed,
+			fmt.Sprintf("failed to execute strategy: %q", err),
+		)
+		diff.Append(releaseutil.SetReleaseCondition(&rel.Status, *releaseStrategyExecutedCond))
+		return rel, nil, nil, err
+	}
```
juliogreff: Similar to 382b0907eb0b1ba0c993f8d320ce39597f7bdb06 in #222, maybe we should move the condition setting to a higher level (both here and in `scheduleRelease`)? Then, all condition setting happens in the same place, and event logging gets combined.

pkg/controller/release/release_controller.go
```diff
@@ -279,39 +268,29 @@ func (c *Controller) syncOneReleaseHandler(key string) error {
...
-	c.applicationWorkqueue.Add(appKey)
-
-	klog.V(4).Infof("Done processing Release %q", key)
-
-	return nil
-}
+	diff := diffutil.NewMultiDiff()
+	defer func() {
+		if !diff.IsEmpty() {
+			// initialRel is kept for the sake of safety: it's
```
juliogreff: `initialRel` has been renamed `baseRel`

pkg/controller/release/release_controller.go
```diff
@@ -382,9 +356,141 @@ func (c *Controller) scheduleRelease(rel *shipper.Release) (*shipper.Release, er
...
+	return releases, nil
+}
+
+func (c *Controller) ensureReleaseState(relinfo *releaseInfo) ([]ExecutorResult, []ReleaseStrategyStateTransition, error) {
+	diff := diffutil.NewMultiDiff()
+	defer func() {
+		if !diff.IsEmpty() {
+			c.recorder.Eventf(relinfo.release, corev1.EventTypeNormal, "ReleaseConditionChanged", diff.String())
+		}
+	}()
```
juliogreff: This diff is no longer used, so it can be deleted.

pkg/controller/release/release_controller.go
```diff
@@ -382,9 +356,141 @@ func (c *Controller) scheduleRelease(rel *shipper.Release) (*shipper.Release, er
 	)
 	diff.Append(releaseutil.SetReleaseCondition(&rel.Status, *condition))
 
-	klog.V(4).Infof("Release %q has been successfully scheduled", controller.MetaKey(rel))
+	strategyPatches, trans, err = c.ensureReleaseState(relinfo)
+	patches = append(patches, strategyPatches...)
```
juliogreff: `strategyPatches` can be `nil` if `err != nil`, so better move this append after the error check.

PR #255, opened 2020-01-07
scheduler: wrap mismatched owner references in a shipper error

juliogreff:
This prevents shipper from making assumptions on whether to retry errors
cause by mismatched owner references for target objects:

	Cannot determine if untagged error &errors.StatusError{...} is
	retriable, will assume it is

This should be a non-retriable error, as it's an unrecoverable condition
for shipper. This happens when a release is deleted and then re-created,
and the correct course of action is for shipper to wait for kubernetes
to garbage collect the old target objects, which will trigger a new sync
of the Release, which can then create the new objects correctly.

### DIFF
--- a/pkg/errors/errors.go
+++ b/pkg/errors/errors.go
@@ -4,8 +4,18 @@ import (
 	"fmt"
 	"strings"
 
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/klog"
 )
+
+// kubeobj is close to an union of Object and Type, as there's no interface in
+// apimachinery that has them both together, and it's used in a few places here
+// to get k8s objects as a parameter.
+type kubeobj interface {
+	GetName() string
+	GetNamespace() string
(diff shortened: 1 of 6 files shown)


pkg/errors/errors.go
```diff
@@ -4,9 +4,19 @@ import (
...
 	"strings"
 
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/klog"
 )
 
+// kubeobj is close to an union of Object and Type, as there's no interface in
+// apimachinery that has them both together, and it's used in a few places here
+// to get k8s objects as a parameter.
+type kubeobj interface {
```
icanhazbroccoli: Should we move it outside errors package and make it public maybe? I see some potential in a broader usage of this interface.

juliogreff: Yes, we probably should, I wanted to reach out for it once or twice. Any ideas in which package this should live?

PR #256, opened 2020-01-08
Better reporting for stuck Deployment

juliogreff:
This builds on top of #242.

One of the bigger challenges for shipper users is to decide if a rollout
is waiting for the right bits to be flipped up in the kubernetes cloud,
or if something went wrong with no hope of being fixed without
intervention. Although this commit does not fundamentally fix that (as
it would be very involved and error prone, requiring the capacity
controller to have intimate details of replica sets and pods, which we
want to avoid), we're now checking for a few more conditions in the
Deployment that surface known situations:

* A Deployment has just been changed, and its Status does not reflect
the brand new Spec. In that case, we considered capacity to be just in
progress.

* A Deployment times out. This is not super common, as it requires users
to define a timeout in the Deployment itself. If that ever happens,
though, we're covered :)

* A Deployment would violate quotas, or would otherwise cause the
ReplicaSet to be in a state of error. That's insanely common, and also
super hard to diagnose without knowing that this is an error condition
to begin with.


pkg/controller/capacity/capacity_controller.go
```diff
@@ -375,82 +447,94 @@ func (c *Controller) subscribeToDeployments(informerFactory kubeinformers.Shared
...
-			ServerError,
-			err.Error(),
-		)
-		diff.Append(capacityutil.SetClusterCapacityCondition(clusterStatus, *cond))
-
 		return nil, err
 	}
 
-	patchString := fmt.Sprintf(`{"spec": {"replicas": %d}}`, replicaCount)
+	patch := []byte(fmt.Sprintf(`{"spec": {"replicas": %d}}`, replicaCount))
```
icanhazbroccoli: Shall we introduce some nice-looking abstraction maybe?

PR #261, opened 2020-01-20
release controller: patch only strategy conditions

juliogreff:
This fixes a race condition where the release controller would update a
Release's condition, then patch over it with the values removed, as they
were just added *after* running the strategy executor.

Now, the strategy executor only patches .Status.Strategy for Releases,
while the controller does a good ol' update to change
.Status.AchievedStep and .Status.Conditions.

### DIFF
--- a/pkg/controller/release/strategy_executor.go
+++ b/pkg/controller/release/strategy_executor.go
@@ -2,35 +2,27 @@ package release
 
 import (
 	"fmt"
-	"sort"
 	"time"
 
-	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
-	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/client-go/tools/record"
 	"k8s.io/klog"
 
 	shipper "github.com/bookingcom/shipper/pkg/apis/shipper/v1alpha1"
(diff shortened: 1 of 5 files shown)


pkg/controller/release/strategy_executor.go
```diff
@@ -2,44 +2,37 @@ package release
...
 
-type PipelineContinuation bool
+type PipelineContinuation int
 
 const (
-	PipelineBreak    PipelineContinuation = false
-	PipelineContinue                      = true
+	PipelineBreak PipelineContinuation = iota
+	PipelineContinue
+	PipelineComplete
```
icanhazbroccoli: I'm a bit concerned about this approach as it might introduce some unwanted coupling between pipeline stages, i.e. steps will know about each other. The initial idea of a regular pipeline step was: can you proceed this state? yes: return pipeline continue; no: return break as it makes no sense to do anything with it. Maybe I over-emphasised the actual meaning of Break, which simply meant to break the loop but didn't mean a negative meaning exclusively.

juliogreff: @icanhazbroccoli I reverted these changes, and instead just made sure that `genReleaseStrategyStateEnforcer` returns `PipelineContinue`. If the last step in a pipeline returns that, it is meant to be success, so that addresses the implicit dependency between stages.

PR #263, opened 2020-01-21
Test Helm repo

hihilla:
Created a dummy helm repo in our github repo.
This means that we do not need to spin a test helm repo when e2e testing.
We can also point to this repo to get the demo application chart for our docs.


Makefile
```diff
@@ -95,20 +94,14 @@ install-shipper: build/shipper.image.$(IMAGE_TAG) build/shipper.deployment.$(IMA
...
 # Run all end-to-end tests. It does all the work necessary to get the current
 # version of shipper on your working directory running in kubernetes, so just
 # running `make -j e2e` should get you up and running immediately. Do remember
 # do setup your clusters with `make setup` though.
-e2e: install install-helm build/e2e.test
+e2e: install build/e2e.test
 	./build/e2e.test --e2e --kubeconfig ~/.kube/config \
-		--testcharts http://$(shell $(KUBECTL) get service helm -o jsonpath='{.spec.clusterIP}'):8879 \
 		--appcluster $(SHIPPER_CLUSTER) \
+	    --testcharts $(TEST_HELM_REPO_URL) \
```
juliogreff: The alignment here doesn't look right. Can't tell from github ui if this is the cause, nut make sure you're using tabs :)

ci/e2e.sh
```diff
@@ -4,7 +4,7 @@
 make setup
 
 # Run the e2e tests, save exit code for later
-DOCKER_REGISTRY=${DOCKER_REGISTRY:=localhost:32000} E2E_FLAGS="--test.v" make -j e2e
+TEST_HELM_REPO_URL=https://raw.githubusercontent.com/bookingcom/shipper/${TRAVIS_COMMIT}/test/e2e/testdata DOCKER_REGISTRY=${DOCKER_REGISTRY:=localhost:32000} E2E_FLAGS="--test.v" make -j e2e
```
juliogreff: Suggested change:

```
TEST_HELM_REPO_URL=${TEST_HELM_REPO_URL:https://raw.githubusercontent.com/bookingcom/shipper/${TRAVIS_COMMIT}/test/e2e/testdata}
```

This makes `TEST_HELM_REPO_URL` use the existing environment variable if it exists, but otherwise use `https://raw.githubusercontent.com/bookingcom/shipper/${TRAVIS_COMMIT}/test/e2e/testdata` as default. That's important for us to overwrite it from inside our internal CI pipeline, so we can point it to whatever chart repo we want.

PR #266, opened 2020-01-22
StrategyExecutor: renamed ExecutorResult to StrategyPatch

icanhazbroccoli:
This commit is a series of cosmetic refactoring changes dedicated to
`StrategyExecutor` inner naming cleanup.

`StrategyPatch` substituted `ExecutorResult` interface and gained 2 more new
methods:
  * `Alters(interface{}) bool`
  * `IsEmpty() bool`

`Alters()` method takes a single argument and checks if applying the patch
against the provided argument object will take any effect. In the future we are planning to shrink down strategy executor's knowledge about internals of target objects and make it work with patches only. Therefore Alters() is a helper method added for now, which could be gone in the future. `Release`, `CapacityTarget` and `TrafficTarget` are already a subject of inner Shipper knowledge and therefore we can cherry-pick the ones that actually introduce changes on the existing objects by running a `DeepEqual` operation against these objects. Assuming some future patch declarations might be disconnected from underlying k8s object definitions, `Alters()` could always return true on these "avatar" objects.

`IsEmpty()` method indicates if the patch contains any declared changes. An
empty patch should always return false on `Alters()` method call.


pkg/controller/release/strategy_patches.go
```diff
@@ -0,0 +1,97 @@
...
+	b, _ := json.Marshal(patch)
+	return c.Name, shipper.SchemeGroupVersion.WithKind("TrafficTarget"), b
+}
+
+func (c *TrafficTargetSpecPatch) Alters(o interface{}) bool {
+	tt, ok := o.(*shipper.TrafficTarget)
+	if !ok || c.IsEmpty() {
+		return false
+	}
+	return !equality.Semantic.DeepEqual(&tt.Spec, c.NewSpec)
```
juliogreff: Given how `checkTraffic` works, I don't think this `DeepEqual` check will ever return true, so I'm skeptical about having it here, both for the small performance penalty, and for the possibility of this hiding bugs in `checkTraffic` itself if it ever starts spewing out no-op patches. The same goes for `CapacityTargetSpecPatch`.

PR #270, opened 2020-01-28
traffic controller: sort pods to shift

juliogreff:
This avoids a (benign?) race condition where we could patch the same
pods several times to have traffic labels, due to unstable order.
Although this eventually converges in the wild, it was causing one of
the tests to flake every once in a while.


icanhazbroccoli:
Nice catch!

PR #274, opened 2020-01-30
ReleaseController detects target step inconsistencies

icanhazbroccoli:
This commit proposes a change to ReleaseController which makes it review
the consistency of the release timeline. The idea is that a release
targetStep can never be less than it's successor's, otherwise this
situation can provoke shipper to reactivate a historical release. This
commit ensures a release in an inconsistent state can not be
reactivated.

This commit might be not the final version and is aiming to spawn a discussion about the desired shipper behavior in a case of inconsistency. During the verbal discussion, we concluded there are 2 major ways: autocorrecting the targetStep or to bail out and expect the user to get it fixed.

### DIFF
--- a/pkg/controller/release/release_controller.go
+++ b/pkg/controller/release/release_controller.go
@@ -420,6 +420,20 @@ func (c *Controller) executeReleaseStrategy(relinfo *releaseInfo, diff *diffutil
 		}
 	}
 	if succ != nil {
+		// If there is a successor to the current release, we have to ensure
+		// it's spec points to the last strategy step.
+		if !releaseutil.IsLastStrategyStep(relinfo.release) {
+			// In practice, this situation most likely means an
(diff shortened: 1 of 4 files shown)


pkg/controller/release/release_controller.go
```diff
@@ -474,6 +474,23 @@ func (c *Controller) ensureReleaseState(relinfo *releaseInfo) (bool, []StrategyP
 		}
 	}
 	if succ != nil {
+		curTargetStep := relinfo.release.Spec.TargetStep
+		succTargetStep := succ.Spec.TargetStep
+
+		if curTargetStep < succTargetStep {
```
juliogreff: Interesting question: what happens when the successor has more strategy steps than the current one (aka, a release that introduces a more complex strategy)?

PR #282, opened 2020-02-07
Release StrategyExecutor aborts if successor release is progressing

icanhazbroccoli:
This release is a fix of a problem with the new implementation of the
reconciliation loop involving strategy execution for all release (as
opposed to apply it to contender-incumbent pair only in the old
implementation). The problem we observed boiled down to the next
scenario:
  1. A contender runs through the strategy execution loop and bails out
  with an incomplete state.
  2. Incumbent enters it's independent strategy loop and checks it's
  desired state from the desired state of the contender. This is the
  point where the things broke: desired state != achieved state. At this
  moment incumbent can start progressing without waiting for it's
  successor (contender) to complete.

This optimistic triggering was not a planned action and was mainly
mistakenly introduced because distinct release generations are being
processed independently now.

Preventing the strategy loop from progressing is one way to get around
this problem where releases are being processed independently. Another
alternative to that could be: an incumbent release can be smarter in
terms of figuring out it's desired state itself (e.g. if it's successor
hasn't achieved it's target step yet, try to reconcile on the last
step in it's own strategy). This behavior has pros and cons of course.
On the bright side: releases can behave much more independently and this
can also challenge the need for incumbent and contender strategy states.
On the lowlight, this might provoke some unwanted chatting and flapping
in the release squad. Chatty releases might be quite expensive in terms
of converging to a stable state (as we now provoke a direct and reverse
chain reaction when there is a change to a release object).

For now, we settle down with an easy patch which prevents the system
from doing unwanted things. This approach might be a subject for change
in the future.

### DIFF
--- a/pkg/controller/release/strategy_executor.go
+++ b/pkg/controller/release/strategy_executor.go
@@ -4,26 +4,24 @@ import (
 	"fmt"
 	"time"
 
-	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/klog"
 
 	shipper "github.com/bookingcom/shipper/pkg/apis/shipper/v1alpha1"
 	"github.com/bookingcom/shipper/pkg/controller"
 	shippererrors "github.com/bookingcom/shipper/pkg/errors"
 	"github.com/bookingcom/shipper/pkg/util/conditions"
+	releaseutil "github.com/bookingcom/shipper/pkg/util/release"
 )
 
 type StrategyExecutor struct {
 	curr, prev, succ *releaseInfo
-	hasIncumbent     bool
 }
 
-func NewStrategyExecutor(curr, prev, succ *releaseInfo, hasIncumbent bool) *StrategyExecutor {
+func NewStrategyExecutor(prev, curr, succ *releaseInfo) *StrategyExecutor {
 	return &StrategyExecutor{
-		curr:         curr,
-		prev:         prev,
-		succ:         succ,
-		hasIncumbent: hasIncumbent,
+		prev: prev,
+		curr: curr,
+		succ: succ,
 	}
 }
 
@@ -34,7 +32,7 @@ const (
 	PipelineContinue                      = true
 )
 
-type PipelineStep func(*StrategyExecutor, conditions.StrategyConditionsMap) (PipelineContinuation, []StrategyPatch, []ReleaseStrategyStateTransition)
+type PipelineStep func(*shipper.RolloutStrategy, int32, Extra, conditions.StrategyConditionsMap) (PipelineContinuation, []StrategyPatch, []ReleaseStrategyStateTransition)
 
 type Pipeline []PipelineStep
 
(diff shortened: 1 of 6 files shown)


pkg/controller/release/strategy_executor.go
```diff
@@ -182,37 +170,21 @@ func genCapacityEnforcer(curr, succ *releaseInfo) PipelineStep {
...
-			strategyStep = strategy.Steps[targetStep]
-			trafficWeight = strategyStep.Traffic.Incumbent
+			trafficWeight = strategy.Steps[targetStep].Traffic.Incumbent
 			condType = shipper.StrategyConditionIncumbentAchievedTraffic
 		}
 
-		isLastStep := int(targetStep) == len(strategy.Steps)-1
-
 		if achieved, newSpec, reason := checkTraffic(curr.trafficTarget, uint32(trafficWeight)); !achieved {
-			e.info("release hasn't achieved traffic yet")
```
juliogreff: Now that `info` is not in use anymore, we can delete the func too

pkg/errors/release.go
```diff
@@ -194,3 +194,24 @@ func NewNotWorkingOnStrategyError(contenderReleaseKey string) NotWorkingOnStrate
...
 	}
 }
+
+type ContenderStrategyIncompleteError struct {
+	relKey string
+}
+
+func (e ContenderStrategyIncompleteError) Error() string {
+	return fmt.Sprintf(
+		"release contender %q strategy is incomplete yet",
```
juliogreff: This reads weird. I'd suggest either "[...] is still incomplete" or "[...] is not complete yet".

pkg/errors/release.go
```diff
@@ -194,3 +194,24 @@ func NewNotWorkingOnStrategyError(contenderReleaseKey string) NotWorkingOnStrate
...
+		e.relKey,
+	)
+}
+
+func (e ContenderStrategyIncompleteError) ShouldRetry() bool {
+	return true
+}
```
juliogreff: I'm not sure if retries here are appropriate. I'd expect that when the contender achieves its desired target step, it would re-enqueue its neighbors, making the blind retries redundant.

pkg/controller/release/strategy_executor.go
```diff
@@ -328,18 +297,20 @@ func genReleaseStrategyStateEnforcer(curr, succ *releaseInfo) PipelineStep {
...
+	// There is no really a point in making any changes until the successor
+	// has completed it's transition, therefore we're hoilding off and aborting
+	// the pipeline execution. An alternative to this approach could be to make
+	// an autonomous move purely based on the picture of the world. But due to
+	// the limited visilibility of what's happening to the successor (as it
+	// might be following it's successor) it could be that a preliminary action
+	// would create more noise than help really.
+	if !isHead {
+		if !releaseutil.ReleaseAchievedTargetStep(e.succ.release) {
+			return false, nil, nil, shippererrors.NewContenderStrategyIncompleteError(
```
juliogreff: Bailing out here does sound like the right thing to do, but is it really an "error" though? It's very much expected behavior that a release will be processed when it's not the head, and its successor hasn't achieved its target step yet: it's basically what every rollout is. Thoughts?

icanhazbroccoli: The error class was mainly introduced for the sake of bubbling up some visibility. I'd agree: there is nothing exceptional happening to the current release if it's successor is still progressing. Removed the error value from return values.

PR #283, opened 2020-02-11
Fixed out-of-range index dereferencing error in release controller

icanhazbroccoli:
This commit fixes a problem where a corresponding release strategy was
resolved twice: once for an actual execution and second time for
reporting. These resolutions were happening in distinct places: the 1st
one in strategy executor, the latter in release controller. As a result,
the release controller one was causing a panic as it was not taking into
account the updated logic of strategy resolution where an incumbent is
supposed to look ahead and use it's successor's strategy, and the index
validity was only happening in strategy executor, which was calculating
the desired strategy correctly. This commit is also moving things
around: strategy executor is being initialized with a specific strategy
and a pointer to the target step and Execute() step takes the sequence
of executable releases as arguments.

### DIFF
--- a/pkg/controller/release/release_controller.go
+++ b/pkg/controller/release/release_controller.go
@@ -273,15 +273,13 @@ func (c *Controller) syncOneReleaseHandler(key string) error {
 	}
 
 	var condition *shipper.ReleaseCondition
-	var strategyPatches []StrategyPatch
-	var trans []ReleaseStrategyStateTransition
 	var relinfo *releaseInfo
-	var stepAchieved bool
+	var patches []StrategyPatch
+	var execRel *shipper.Release
 
 	// we keep baseRel as a comparison baseline in order to figure out if
 	// we even have to send an update
 	baseRel := rel.DeepCopy()
-	patches := make([]StrategyPatch, 0)
 
 	diff := diffutil.NewMultiDiff()
 	defer func() {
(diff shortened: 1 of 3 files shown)


pkg/controller/release/release_controller.go
```diff
@@ -454,41 +400,122 @@ func (c *Controller) applicationReleases(rel *shipper.Release) ([]*shipper.Relea
...
+		var achievedStepName string
+		if isHead {
+			achievedStep = targetStep
+			achievedStepName = strategy.Steps[achievedStep].Name
+		} else {
+			achievedStep = int32(len(rel.Spec.Environment.Strategy.Steps)) - 1
+			achievedStepName = rel.Spec.Environment.Strategy.Steps[achievedStep].Name
+
+		}
+		if prevStep == nil || achievedStep != prevStep.Step {
```
hihilla: I have a small concern: we are trusting the pipeline to not be `complete` when the target step was not achieved, so if we arrived here, it means 100% target step was achieved and achievedStep == targetStep?

icanhazbroccoli: It's quite confusing indeed. Achieved step is the one we're reporting. targetStep might be different as it's the one that comes from the successor (if any) strategy. If the current release is not head, here in this very place we simply ensure that it's state reports the last step in it's own strategy, no in it's successor which is being used as a desired state.

PR #285, opened 2020-02-19
Bugfixes in release controller strategy executor

icanhazbroccoli:
This commit addresses multiple issues identified in strategy executor,
among which:
  * Wrong recepients for patch updates: due to an error in the code some
  patches were applied to a wrong generation of release and target
  objects.
  * Target object spec checkers used to return an incomplete spec if
  only some of the clusters are misbehaving: there was a risk of
  de-scheduling the workload on healthy clusters.

### DIFF
--- a/pkg/controller/release/strategy_executor.go
+++ b/pkg/controller/release/strategy_executor.go
@@ -32,14 +32,15 @@ func (p *Pipeline) Enqueue(step PipelineStep) {
 }
 
 type Extra struct {
-	IsLastStep   bool
-	HasIncumbent bool
+	IsLastStep bool
+	HasTail    bool
+	Initiator  *shipper.Release
 }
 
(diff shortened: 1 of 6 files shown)


pkg/controller/release/strategy_executor.go
```diff
@@ -32,14 +32,15 @@ func (p *Pipeline) Enqueue(step PipelineStep) {
 }
 
 type Extra struct {
-	IsLastStep   bool
-	HasIncumbent bool
+	IsLastStep bool
+	HasTail    bool
+	Initiator  *shipper.Release
```
juliogreff: I get what the Initiator is after reading the code, but without being familiar with how things work I don't think that the name carries enough meaning. I'm sorry I don't have a better suggestion, so I'll still approve, and if a better name comes up we can have a PR just to rename this.

PR #287, opened 2020-02-21
Surface CapacityTarget issues in the release

juliogreff:
This looks like it doesn't change anything in the Release, but that's mostly because the Release was just reading the CapacityTarget conditions anyway :)

Before this, the capacity controller would only put a list of unready
clusters in the CapacityTarget's Ready condition when it would set it to
False. This requires users to go digging into each cluster condition,
and most likely they would only be directed to SadPods, where they could
finally get some useful information.

Now, that information is summarized in a very brief format, in the hopes
that users will have to do less jumping around when investigating why
their CapacityTarget is not progressing.

For instance, if the CapacityTarget is stuck because one container can't
pull its image, we'll now have the following in the CapacityTarget's
.stauts.conditions:

```
[
  {
    "lastTransitionTime": "2020-02-12T13:16:44Z",
    "status": "True",
    "type": "Operational"
  },
  {
    "lastTransitionTime": "2020-02-12T13:16:44Z",
    "message": "docker-desktop: PodsNotReady 3/3: 3x\"test-nginx\" containers with [ImagePullBackOff]",
    "reason": "ClustersNotReady",
    "status": "False",
    "type": "Ready"
  }
]
```

As a bonus, this is also shown on a `kubectl get ct`:

```
% kubectl get ct  snowflake-db84be2b-0
NAME                   OPERATIONAL   READY   REASON                                                                                AGE
snowflake-db84be2b-0   True          False   docker-desktop: PodsNotReady 3/3: 3x"test-nginx" containers with [ImagePullBackOff]   8d
```
This addresses the CT part of both #139 and #79.

### DIFF
--- a/pkg/controller/capacity/deployment.go
+++ b/pkg/controller/capacity/deployment.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"math"
 	"sort"
+	"strings"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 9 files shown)


pkg/controller/capacity/deployment.go
```diff
@@ -94,3 +95,62 @@ func (c Controller) calculatePercentageFromAmount(total, amount int32) int32 {
...
+	return strings.Join(summaryStrs, "; ")
+}
+
+func summarizeContainers(containers []corev1.ContainerStatus, sadContainerReasons map[string]*sadContainerSummary) {
+	for _, container := range containers {
+		if container.Ready {
+			continue
+		}
+
+		if sadContainerReasons[container.Name] == nil {
```
icanhazbroccoli: Would you please elaborate what this case means? At first it looked like a miss in the hashtable lookup, but i see it's a nil-check. Where do nils are supposed to come from?

juliogreff: It *should* check for a lookup miss. Thanks for spotting!

PR #290, opened 2020-03-02
Clustrerclientstore embeds shipper clientset and informer factory refs #276

### LINKED ISSUE #276: Add a shipper clientset and informer factory to clusterclientstore
This is a mostly independent, non-breaking, supporting task to allow the release controller to operate on target objects in application clusters. Our current clusterclientstore (a component holding clients to each application cluster) only has  `GetClient(string, string) (kubernetes.Interface, error)` and `GetInformerFactory(string) (kubeinformers.SharedInformerFactory, error)` methods, but we’ll also need `shipperclientset.Interface` and `shipperinformers.SharedInformerFactory` as well.

Since we’re touching this component anyway, I’d also like to suggest a few changes to the interfaces here. One major gripe I have with the current version can be illustrated [by this piece of code](https://github.com/bookingcom/shipper/blob/92341210587b7f2685b8d46b0c6c32a8ef87b9c0/pkg/controller/installation/installation_controller.go#L444-L460):

```
func (c *Controller) GetClusterAndConfig(clusterName string) (kubernetes.Interface, *rest.Config, error) {
	client, err := c.clusterClientStore.GetClient(clusterName, AgentName)
	if err != nil {
		return nil, nil, err
	}

	referenceConfig, err := c.clusterClientStore.GetConfig(clusterName)
	if err != nil {
		return nil, nil, err
	}

	referenceCopy := rest.CopyConfig(referenceConfig)

	return client, referenceCopy, nil
}
```

A client, a config and an informer factory all either exist or don’t exist for a given `clusterName`, but yet we treat them all as separate entities. I propose the following interfaces:

```
package clusterclientstore

type Interface interface {
    // these two remain unchanged:
    AddEventHandlerCallback(EventHandlerRegisterFunc)
    AddSubscriptionCallback(SubscriptionRegisterFunc)

    GetApplicationClusterClientsets(clusterName, ua string) (ClusterClientsets, error)
}

type ClusterClientsets interface {
    GetConfig() *rest.Config

    GetKubeClient() kubernetes.Interface
    GetKubeInformerFactory() kubeinformers.SharedInformerFactory
    
    GetShipperClient() shipperclientset.Interface
    GetShipperInformerFactory() shipperinformers.SharedInformerFactory
}
```

This should simplify a lot of error handling and argument passing in the controllers.

icanhazbroccoli:
This commit changes the signature of clusterclientstore interface and
    arms it with `GetApplicationClusterClientset/2` method which returns an
    instance of `clusterclientstore.ClientsetInterface`. The new composite
    interface is supposed to embed kubernetes and shipper clientsets and
    informer factories. The aforementioned method is the one that can return
    an error, which is opposite to the former `clusterclientset.Interface`
    behavior where every independent method call to `GetConfig/1`,
    `GetClient/1` and `GetInformerFactory/1` could return an error
    independently, leaving the client with a responsibility if it's ok to
    proceed forward.

Fixes #276

### DIFF
--- a/cmd/shipper-app/main.go
+++ b/cmd/shipper-app/main.go
@@ -31,6 +31,7 @@ import (
 	shipper "github.com/bookingcom/shipper/pkg/apis/shipper/v1alpha1"
 	"github.com/bookingcom/shipper/pkg/chart/repo"
 	"github.com/bookingcom/shipper/pkg/client"
+	shipperclientset "github.com/bookingcom/shipper/pkg/client/clientset/versioned"
 	shipperscheme "github.com/bookingcom/shipper/pkg/client/clientset/versioned/scheme"
 	shipperinformers "github.com/bookingcom/shipper/pkg/client/informers/externalversions"
 	"github.com/bookingcom/shipper/pkg/clusterclientstore"
@@ -104,7 +105,7 @@ func main() {
 	klog.InitFlags(nil)
 	flag.Parse()
 
-	baseRestCfg, err := clientcmd.BuildConfigFromFlags(*masterURL, *kubeconfig)
+	restCfg, err := prepareRestConfig()
 	if err != nil {
 		klog.Fatal(err)
 	}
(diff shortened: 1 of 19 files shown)


icanhazbroccoli:
@juliogreff this is a draft version of this PR and I'd be super happy to hear your thoughts and suggestions. Thanks!

juliogreff:
I can take another harder look and nitpick, but so far LGTM. What's needed to take it out of draft?

icanhazbroccoli:
@juliogreff I mainly wanted to make sure I followed your guidance correctly and left no elephants in the room. Commit management is the only todo so far.

Issue #277, opened 2020-01-30
Move target object to application clusters


juliogreff:
Currently, all CRDs are defined only in the management cluster. As part of tackling #272, we should instead define the `InstallationTarget`, `CapacityTarget` and `TrafficTarget` CRDs in the application clusters. This is where the actual split of Shipper’s functionality will start, and probably where most of the work will end up. Given that this will result in a massive diff, I suggest that we split the work required to resolve this issue into 3 separate pull requests, one for each target object. This will probably allow us to better learn about how the end result should look like through shorter review cycles.

A non-exhaustive list of things to be considered:

* Change `shipperctl` to install target object CRDs in the application clusters instead of the management cluster. We might need to change the app cluster Roles and ClusterRoles to allow Shipper to operate on them as well.
* Move all target object properties from `.spec.clusters` directly into `.spec`, as each target object is now supposed to represent the desired state of a single application cluster.
* Remove `.status.clusters` from the target objects, and use the top level conditions at `.status.conditions` to represent the status of the cluster.
* Change the release controller to operate on target objects in the application clusters. This requires the implementation of #276.

PR #298, opened 2020-03-13
Assorted Refactorings

juliogreff:
While working on #277, a few small refactorings made life easier. Since the diff for that issue is getting fairly large, I'm backporting everything I can into this PR, to make reviewing the actual moving of CRDs easier when it does land.


cmd/shipper-app/main.go
```diff
@@ -348,19 +348,8 @@ func startInstallationController(cfg *cfg) (bool, error) {
 		return false, nil
 	}
 
-	dynamicClientBuilderFunc := func(gvk *schema.GroupVersionKind, config *rest.Config, cluster *shipper.Cluster) dynamic.Interface {
-		config.APIPath = dynamic.LegacyAPIPathResolverFunc(*gvk)
-		config.GroupVersion = &schema.GroupVersion{Group: gvk.Group, Version: gvk.Version}
-
-		if cfg.restTimeout != nil {
```
hihilla: Are we not copying the timeout anymore?

juliogreff: We are, but now just once. The `restConfig` this `dynamicClientBuilderFunc` gets comes from the `clusterclientstore` (by means of its `GetConfig` method), and the timeout is already set here: https://github.com/bookingcom/shipper/blob/master/pkg/clusterclientstore/store.go#L304-L308

PR #305, opened 2020-03-31
release controller: refactor tests

juliogreff:
This depends on changes in #304, so let's get that reviewed before this one!

Previously, the release controller tests relied way too much on the
actions it took to achieve a particular state, instead of just checking
the final state directly. Here, we're taking the same approach that we
took with the other controllers: check just the final state, without
bothering too much about the actions we took to get there, as it makes
for less brittle tests, and refactoring much easier.

You'll also notice that a bunch of test cases are gone. Those are the
ones that were testing behavior of the scheduler and the strategy
executor. Scheduler tests are good enough, but you'll notice that the
strategy executor is particularly devoid of them. That's a compromise
I'm making to get this out the door quickly, but we'll coordinate to
introduce tests to the strategy executor later.

There's also a small opportunistic refactor on package client that I included here just to avoid opening yet another PR.

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -3,13 +3,16 @@ package client
 import (
 	"fmt"
 	"runtime"
+	"strconv"
 
+	corev1 "k8s.io/api/core/v1"
 	apiextensionclientset "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 
+	shipper "github.com/bookingcom/shipper/pkg/apis/shipper/v1alpha1"
 	shipperclientset "github.com/bookingcom/shipper/pkg/client/clientset/versioned"
 	"github.com/bookingcom/shipper/pkg/version"
 )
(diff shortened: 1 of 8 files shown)


pkg/client/client.go
```diff
@@ -34,6 +37,44 @@ func decorateConfig(config *rest.Config, ua string) *rest.Config {
...
+		config.CAData = ca
+
+		token := secret.Data["token"]
+		config.BearerToken = string(token)
+		return config
+	}
+
+	// Let's figure it's either a TLS secret or an opaque thing formatted
+	// like a TLS secret.
+	if ca, ok := secret.Data["tls.ca"]; ok {
```
icanhazbroccoli, line 57: CA bundle comes in base64 encoding. I wonder if we have to do an explicit decoding before feeding this to the config?

juliogreff: From the docs:

> CAData holds PEM-encoded bytes (typically read from a root certificates bundle).

Further, it seems like PEM-encoded bytes essentially means base64 (https://golang.org/src/encoding/pem/pem.go):

```
// The encoded form is:
//    -----BEGIN Type-----
//    Headers
//    base64-encoded Bytes
//    -----END Type-----
// where Headers is a possibly empty sequence of Key: Value lines.
```

So I read that this means no decoding is necessary, the CAData is expected to be in base64 already.

icanhazbroccoli: Great, thanks for the clarification!

PR #306, opened 2020-04-06
Move InstallationTargets to application clusters

juliogreff:
Currently, all CRDs are defined only in the management cluster. As part
instead defining the InstallationTarget CRD in the application clusters.
This is where the actual split of Shipper’s functionality starts. This
only splits InstallationTargets because of the massive diff that's being
generated, so I hope this will be easier to review.

As a general warning, you'll see a lof of weird scaffolding code for the
moment, like passing both management clients and clusterclientstore
clients alongside each other, installing shipper-app in *both* clusters,
and others. That should all go away once we move the other target
objects as well. Review those gently :)

This also completely obliterates the janitor controller. It just won't
work in this setup anymore, and needs a complete rewrite. That will come
as a follow up to this commit (again, to make diffs smaller).


cmd/shipperctl/cmd/clusters.go
```diff
@@ -111,6 +120,44 @@ func runSetupMgmtClusterCommand(cmd *cobra.Command, args []string) error {
...
+
+	for _, cluster := range clusterList.Items {
+		cmd.Printf("Setting up application clusters %s:\n", cluster.Name)
+
+		secret, err := mgmtConfigurator.FetchSecret(cluster.Name, shipperNamespace)
+		if err != nil {
+			return err
+		}
+
+		restConfig := shipperclient.BuildConfigFromClusterAndSecret(&cluster, secret)
```
icanhazbroccoli, line 144: Minor: shall we call this var something like: `clusterConfig` instead?

pkg/apis/shipper/v1alpha1/types.go
```diff
@@ -206,6 +206,7 @@ type AchievedStep struct {
 type ReleaseConditionType string
 
 const (
+	ReleaseConditionTypeClustersChosen   ReleaseConditionType = "ClustersChosen"
```
icanhazbroccoli, line 209: I think we never really distinguished 2 states of a release object: clusters chosen and scheduled. Does it make sense to keep "scheduled" around?

juliogreff: "Scheduled" still means that the target objects were created. We definitely need a new name, but I don't think that the actual condition would go away, unless we move it to be inside the actual strategy, but that's way too much work for the context of this PR. I'd happily rename it though, I just haven't figured out a better name yet.

pkg/controller/installation/installation_controller_test.go
```diff
@@ -2,220 +2,158 @@ package installation
...
-				status:             status,
-			},
-		},
-	)
+	runInstallationControllerTest(t, it, status, nil)
 }
 
 // buildExpectedObjects returns a list of the objects we expect from
-// `chartName`. This can be hardcoded for as long as we depend on that one chart.
+// `nginxChartName`. This can be hardcoded for as long as we depend on that one
```
icanhazbroccoli, line 66: Eventually it should become a kubernetes core api constant :-D

pkg/apis/shipper/v1alpha1/types.go
icanhazbroccoli: I've finished reading and your words totally make sense to me. Naming is what confused me. Thanks.

PR #308, opened 2020-04-07
WIP: Hilla/gradual rollout v2

### LINKED ISSUE #124: Rolling capacity upgrade on rollouts
When doing a rollout, currently there's a requirement for additional resources available to transition from one step to the next: 
 - _capacity must equal the total incumbent from `current_step - 1` + total contender from `current_step`_

When you have a limited resource quota, this means that you cannot fully take advantage of it. In the worst case, you need to reserve 50% of your resources just to be able to do a rollout. 

Can we please implement an option of transitioning from one rollout step to the next that can cap the number of additional pods used at any given time? 

For example, I would like to tell shipper that it can use up to 2 pods at the time, to make the shift from step A to step B, so that I know I only need to reserve capacity for 2 pods for this.

hihilla:
This is a wip

Will close #124


pkg/controller/release/scheduler.go
```diff
@@ -59,6 +60,19 @@ func (s *Scheduler) ScheduleRelease(rel *shipper.Release) (*releaseInfo, error)
...
 
 	releaseErrors := shippererrors.NewMultiError()
+	if rel.Spec.VirtualStrategy == nil {
+		strategy := rel.Spec.Environment.Strategy
+
+		surgeWeight, err := releaseutil.GetMaxSurgePercent(rel.Spec.Environment.Strategy.RollingUpdate, replicaCount)
+		if err != nil {
+			releaseErrors.Append(err)
+		}
+		virtualStrategy, err := releaseutil.BuildVirtualStrategy(strategy, surgeWeight)
```
juliogreff, line 70: This should probably not execute if there is an error. Same goes for line 74 below.

pkg/util/release/environment_test.go
```diff
@@ -0,0 +1,99 @@
...
+			"empty achieved step",
+			&shipper.AchievedStep{},
+			0,
+			0,
+		},
+		{
+			"achieved step",
+			&shipper.AchievedStep{
+				Step: 0,
+				Name: "0",
```
juliogreff, line 77: This is very nitpicky, but I'd have omitted the Name in the struct, since it's not used. Using it kinda denotes the intention that it *would* be used, but it's not the case here, so I personally find it distracting.

pkg/util/release/max_surge.go
```diff
@@ -0,0 +1,33 @@
...
+func GetMaxSurgePercent(rollingUpdate *shipper.RollingUpdate, replicaCount int32) (int, error) {
+	maxSurgeDefaultValue := intstrutil.FromString("100%")
+	maxSurgeValue := intstrutil.FromString("100%")
+	if rollingUpdate != nil {
+		maxSurgeValue = *intstrutil.ValueOrDefault(&rollingUpdate.MaxSurge, maxSurgeValue)
+	}
+
+	surge, err := intstrutil.GetValueFromIntOrPercent(&maxSurgeValue, int(replicaCount), true)
+	if err != nil {
+		return 100, err
```
juliogreff, line 20: Not that it matters a whole bunch since one is not supposed to use the returned value in case of error, but It's idiomatic to return a zero-value in that case.

pkg/util/release/max_surge.go
```diff
@@ -0,0 +1,33 @@
...
+		// 0 or less are invalid values, using default
+		surge, err = intstrutil.GetValueFromIntOrPercent(&maxSurgeDefaultValue, int(replicaCount), true)
+		if err != nil {
+			return 100, err
+		}
+	}
+	surgePercent := math.Ceil((float64(surge) / float64(replicaCount)) * 100)
+	// surgePercent must be between 1 - 100
+	surgePercent = math.Min(surgePercent, 100)
+	return int(surgePercent), err
```
juliogreff, line 32: Returning an explicit `nil` if there are no errors is probably more explicit, and less bug prone in the long run, since `err` tends to get reused a lot.

pkg/util/release/virtual_strategy.go
```diff
@@ -0,0 +1,101 @@
...
+	virtualStepsIncumbent := math.Ceil(diffIncumbentCapacity / float64(surgeWeight))
+
+	virtualSteps := math.Max(virtualStepsContender, virtualStepsIncumbent)
+	virtualSteps = math.Max(virtualSteps, 1)
+
+	// Calculate traffic surge to each step
+	diffContenderTraffic := float64(nextContenderTraffic - currContenderTraffic)
+	diffIncumbentTraffic := float64(nextIncumbentTraffic - currIncumbentTraffic)
+	trafficSurgeContender := math.Ceil(diffContenderTraffic / virtualSteps)
+	trafficSurgeIncumvent := math.Ceil(diffIncumbentTraffic / virtualSteps)
```
juliogreff, line 66: s/trafficSurgeIncumvent/trafficSurgeIncumbent/ :)

pkg/util/release/virtual_strategy_test.go
```diff
@@ -0,0 +1,260 @@
...
+						Traffic:  shipper.RolloutStrategyStepValue{Incumbent: 0, Contender: 100},
+					},
+				},
+			},
+			&shipper.RolloutVirtualStrategy{
+				Steps: []shipper.RolloutStrategyVirtualStep{
+					{
+						VirtualSteps: []shipper.RolloutStrategyStep{
+							{
+								Name: "0:  to staging",
```
juliogreff, line 44: I wonder, why do we care about giving virtual strategy steps a name? This should be mostly an internal concern, and I don't expect it to aid debugging. Thoughts?

hihilla: I tend to agree, but when debugging I looked both at the numbers and at the title
after some more refactoring I will test it again and see if the title actually helps me

PR #314, opened 2020-04-22
Move CapacityTargets to application clusters

juliogreff:
Currently, most CRDs are defined only in the management cluster. As part
instead defining the CapacityTarget CRD in the application clusters,
together with InstallationTargets, and soon to be joined by
TrafficTargets too.


pkg/controller/capacity/capacity_controller.go
```diff
@@ -41,64 +39,83 @@ const (
...
+
+	recorder record.EventRecorder
 }
 
 // NewController returns a new CapacityTarget controller.
 func NewController(
-	clientset shipperclient.Interface,
+	kubeClient kubernetes.Interface,
+	kubeInformerFactory kubeinformers.SharedInformerFactory,
+	shipperClient shipperclient.Interface,
```
icanhazbroccoli, line 73: Would an object conforming to `ClientsetInterface` be helpful here?

juliogreff: As matter of fact, it would. Thanks for the suggestion! I'll apply it together with the upcoming changes to the traffic controller.

pkg/controller/capacity/capacity_controller.go
```diff
@@ -41,64 +39,83 @@ const (
...
+	shipperClient shipperclient.Interface
+	kubeClient    kubernetes.Interface
 
 	capacityTargetsLister listers.CapacityTargetLister
 	capacityTargetsSynced cache.InformerSynced
 
-	releasesLister       listers.ReleaseLister
-	releasesListerSynced cache.InformerSynced
+	deploymentsLister appslisters.DeploymentLister
+	deploymentsSynced cache.InformerSynced
```
icanhazbroccoli, line 59: Do we care about `ReplicaSets` here?

juliogreff: No, we only directly interact with deployments. Any interesting changes to the ReplicaSet would reflect on the Deployment anyway, so just watching them is sufficient.

PR #315, opened 2020-04-29
Gradual capacity and traffic shifting

### LINKED ISSUE #124: Rolling capacity upgrade on rollouts

hihilla:
Shipper requires extra resources in order to perform a rollout. In case of an application with a staging-vanguard-full on strategy, Shipper will require ~twice as many resources to roll out an application. When having limited resources, this means users cannot fully take advantage of their resources. This aims to provide an answer to this complaint, by progressing in the rollout strategy in smaller increments.

This closes #124


pkg/util/release/virtual_strategy.go
```diff
@@ -0,0 +1,98 @@
...
+			Incumbent: 100,
+			Contender: 0,
+		},
+	}
+	nextStep := strategy.Steps[0]
+	rolloutStrategyVirtualSteps := buildRolloutStrategyVirtualSteps(step, nextStep, surgeWeight)
+	rolloutVirtualStrategy[0] = shipper.RolloutStrategyVirtualStep{VirtualSteps: rolloutStrategyVirtualSteps}
+
+	for i, step := range strategy.Steps {
+		if i != len(strategy.Steps)-1 {
```
juliogreff, line 28: I wonder if this could be better expressed as a range on `strategy.Steps[:len(strategy.Steps)-1]` or similar instead.

pkg/controller/release/release_controller.go
```diff
@@ -392,18 +433,42 @@ func (c *Controller) executeStrategyOnClusters(
...
+	targetStep := releaseutil.GetTargetStep(rel, succ, strategy, stepsFromAchievedStep, progressing, hasAchievedTargetStep)
+	if targetStep >= int32(len(strategy.Steps)) {
+		return rel, shippererrors.NewUnrecoverableError(
+			fmt.Errorf("no step %d in strategy", targetStep))
+	}
+
+	if virtualStrategy == nil {
+		// wait for virtual strategy to be built
+		err := fmt.Errorf("release %q missing a virtual strategy", objectutil.MetaKey(rel))
+		return rel, shippererrors.NewRecoverableError(err)
```
juliogreff, line 466: This should be an unrecoverable error, as the virtual strategy eventually appearing will trigger a new event. Otherwise, shipper will keep on spinning on this until the virtual strategy appears, if ever.

pkg/controller/release/release_controller_test.go
```diff
@@ -388,6 +404,118 @@ func TestLastStep(t *testing.T) {
...
+	// intermediate step
+	rel.Spec.TargetStep = targetStep
+
+	it, tt, ct := buildAssociatedObjectsWithStatus(rel, []*shipper.Cluster{cluster}, &achievedStep)
+
+	// NOTE(jgreff): you'll note that it is in app cluster while ct and tt
+	// are in mgmt. that's scaffolding that needs to remain in place until
+	// we migrate all the objects to the app cluster.
+	mgmtClusterObjects := []runtime.Object{rel, cluster, tt}
+	appClusterObjects := map[string][]runtime.Object{
```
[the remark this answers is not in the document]
hihilla: will do once we merge your branch :)

PR #316, opened 2020-04-29
Move TrafficTargets to application clusters

### LINKED ISSUE #272: Split shipper into a management and an application cluster component
Currently, Shipper cannot tolerate the loss of communication between the management cluster the application clusters. Besides the obvious implication of ongoing Releases no longer being able to progress and complete, which one could say is an acceptable degradation, there are bigger risks outside the setup and teardown of a release, but during an application’s runtime.

As Shipper can no longer enforce the state of Installation, Capacity and Traffic targets, the longer a network partition persists, the greater the likelihood that the actual state of an Application in a cluster is different from the desired state as specified by those targets. This means that objects that get deleted won’t be reinstalled, deployments that get rescaled won’t be scaled back to the desired capacity, and Pod churn will eventually cause the Application’s Service to not target any Pods, as this is also enforced by the traffic controller during runtime.

Our proposed solution is to split Shipper into two components. The first component, let’s call it `shipper-mgmt` at least for the purposes of this issue as this is definitely not set in stone, will remain in the so called management cluster, and will be responsible for being the single source of truth for Applications and Releases. The second component, `shipper-app`, will live in each of the application clusters, and will be the one enforcing the desired state of Installation, Capacity and Traffic targets.

There are several details to be worked out in such a fundamental change, and that’s outside the scope of this issue. This will be used to track the progress of this epic. More details on the work that needs to be done can be found in the following issues:

- [x] #271: Replace microk8s with kind for running end-to-end tests
- [x] #273: Split shipper command into two separate binaries
- [x] #276: Add a shipper clientset and informer factory to clusterclientstore
- [ ] #277: Move target object to application clusters
- [x] #278: Change garbage collection of objects

### LINKED ISSUE #277: Move target object to application clusters

juliogreff:
Currently, most CRDs are defined only in the management cluster. As part
instead defining the TrafficTarget CRD in the application clusters,
together with InstallationTargets and CapacityTargets.

Most of the changes here are inspired by
cbd0fb1674cbcfaedeb02f92034c9e97d04a65f0, so if you've seen that, you
have a pretty good idea of what to expect here :)

Closes #272. Closes #277.


cmd/shipper-app/main.go
```diff
@@ -404,9 +404,10 @@ func startTrafficController(cfg *cfg) (bool, error) {
 	}
 
 	c := traffic.NewController(
-		client.NewShipperClientOrDie(traffic.AgentName, cfg.restCfg),
+		client.NewKubeClientOrDie(capacity.AgentName, cfg.restCfg),
+		cfg.kubeInformerFactory,
+		client.NewShipperClientOrDie(capacity.AgentName, cfg.restCfg),
```
hihilla, line 409: I'm a bit confused. If this is `startTrafficController`, why would it create a client the capacity agent name?

juliogreff: Whoops, I guess? :)

PR #317, opened 2020-05-01
e2e tests fixes

hihilla:
* ApiMaster in clusters object will not be the same as the server url in kubeconfig when using kind to setup an application cluster and a management cluster. This is because the server url in kubeconfig allows the user to communicate with the clusters, but the management cluster requires the url stored in the apiMaster of the clusters object. The e2e tests are behaving as a user when sending requests to the cluster, so we're fetching the server url from kubeconfig and replacing the app clusters apiMaster with it.
* Adding app cluster to teardownNamespace and purgeTestNamespaces.
* Handling errors the moment they happen.

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -13,6 +13,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/klog"
 
+	"github.com/bookingcom/shipper/cmd/shipperctl/configurator"
 	shipper "github.com/bookingcom/shipper/pkg/apis/shipper/v1alpha1"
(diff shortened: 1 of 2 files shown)


test/e2e/e2e_test.go
```diff
@@ -93,6 +95,17 @@ func TestMain(m *testing.M) {
 
 		testRegion = appCluster.Spec.Region
 
+		if *appClusterFromKubeConfig {
+			kubeCluster, err := configurator.NewClusterConfiguratorFromKubeConfig(*kubeconfig, *appClusterName)
+			if err != nil {
+				if err != nil {
```
amosipov, line 101: Checking for the error twice here :)

PR #343, opened 2020-09-09
Fixes for release 0.9

hihilla:
Fix the migration code to not bail out on the first error. The migration code should collect errors on the way. This will help it deal with non-panic errors such as decommissioned cluster better. Making sure the migration code does not try and fail to update an initial target object for a dev cluster.

Fix wrong garbage collection. `SearchStrings` will return the index of the object in the array if found, or the index to insert this object of not found.
It will return 0 if the item is located in index 0 in the array, and it will never return -1.
As we would like to remove orphaned objects for Releases where they are *not* scheduled, we need a better way to check if this cluster is selected.

### DIFF
--- a/pkg/controller/release/migration.go
+++ b/pkg/controller/release/migration.go
@@ -14,22 +14,26 @@ import (
 func (c *Controller) migrateTargetObjects(relName, namespace string) error {
 	selector := labels.Set{shipper.ReleaseLabel: relName}.AsSelector()
 
+	migrationErrors := shippererrors.NewMultiError()
 	err := c.migrateCapacityTargets(relName, namespace, selector)
 	if err != nil {
-		return err
+		//return err
+		migrationErrors.Append(err)
 	}
 
 	err = c.migrateTrafficTargets(relName, namespace, selector)
 	if err != nil {
-		return err
+		//return err
+		migrationErrors.Append(err)
 	}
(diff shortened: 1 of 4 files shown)


ArchangelX360 (approved):

pkg/controller/release/migration.go
```diff
@@ -116,10 +116,19 @@ func (c *Controller) migrateCapacityTargets(relName, namespace string, selector
...
-			return err
+			ctErrors.Append(err)
+		}
+	}
+	if ctErrors.Flatten() == nil && shouldTag {
+		// update initial object with migrated label
+		initialCt.Labels[shipper.MigrationLabel] = "true"
+		_, err = c.clientset.ShipperV1alpha1().CapacityTargets(namespace).Update(initialCt)
+		if err != nil {
+			klog.Error(err)
```
nodo, line 127: nit: should we something before logging the error to make it clearer? E.g. `klog.Errorf("error updating capacity targets %v": err)`

If you decided to change it, I would change it also for `TrafficTargets` and `InstallationTargets`

nodo (commented):
Just a nit, otherwise lgtm 🎉

nodo (approved):

parhamdoustdar (approved):

PR #344, opened 2020-09-09
Shipperctl commands to help with decommissioned clusters

hihilla:
shipperctl commands to clean decommissioned clusters from release annotation, to help shipper forget about these clusters

### DIFF
--- a/cmd/shipperctl/cmd/clean.go
+++ b/cmd/shipperctl/cmd/clean.go
@@ -24,8 +24,8 @@ var (
 		Short: "clean Shipper objects",
 	}
 
-	cleanRelCmd = &cobra.Command{
-		Use:   "releases",
+	cleanDeadClustersCmd = &cobra.Command{
+		Use:   "decommissioned-clusters",
 		Short: "clean Shipper releases from decommissioned clusters",
 		RunE:  runCleanCommand,
 	}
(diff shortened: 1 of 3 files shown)


cmd/shipperctl/cmd/clean.go
```diff
@@ -0,0 +1,238 @@
...
+	countReleasesCmd = &cobra.Command{
+		Use:   "release",
+		Short: "count Shipper *releases* that are scheduled *only* on decommissioned clusters",
+		RunE:  runCountReleasesCommand,
+	}
+)
+
+func init() {
+	// Flags common to all commands under `shipperctl clusters`
+	for _, cmd := range []*cobra.Command{countReleasesCmd, countContendersCmd, cleanDeadClustersCmd} {
```
nodo, line 53: I would use another name here, it's not a big deal but it's the same name as the package :-) so it's a bit confusing.

cmd/shipperctl/cmd/clean.go
```diff
@@ -0,0 +1,238 @@
+package cmd
```
nodo, line 1: nit: I would rename the filename to something like `decommission.go` since it has both "clean" and "count" commands.

parhamdoustdar (approved):

PR #347, opened 2020-09-14
Shipperctl count command output option

hihilla:
This pr adds to the new count command an output option. This allows the users of this command to also get a list of releases for future actions.
For example `shipperctl count contender --decommissionedClusters="decommissioned-a,decommissioned-b"  -o json` 
will output a json of  the form
`[{"namespace": "yfouquet","name": "reserve-backend-test-359dc193-0"},{"namespace": "yfouquet", "name": "reserve-backend-test-359dc193-0" },...]` 
which allows the user to handle these cases separately.

### DIFF
--- a/cmd/shipperctl/cmd/count.go
+++ b/cmd/shipperctl/cmd/count.go
@@ -1,0 +1,182 @@
+package cmd
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/ghodss/yaml"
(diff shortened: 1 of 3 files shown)


nodo:
LGTM 👍

nodo:
OT comment:

Do these commands make sense for non-decommissioned clusters? If this is the case we could change the flag to `--clusters`. WDYT?

```
shipperctl count contender --clusters="..."
```

nodo (dismissed):

nodo (approved):

PR #362, opened 2020-10-13
Ops scripts

hihilla:
#321 

> introducing ops script to spin kind clusters and install shipper on them


ArchangelX360 (approved):

ArchangelX360 (changes requested):
nit: could this be renamed to local dev script instead? This is not meant to be used by operators but by developers.

ArchangelX360 (approved):

PR #372, opened 2020-11-16
shipperctl backup commands

hihilla:
Prepare - saves applications and releases objects to file in a tree (map) form: for each application we store the releases belonging to it.

Restore - restores from a backup file. First restore the application, populate the new UID of the application in its releases owner reference and then apply releases.

output example:
```
$ shipperctl backup -h
Backup and restore Shipper applications and releases

Usage:
  shipperctl backup [command]

Available Commands:
  prepare     Get yaml of application and release objects prepared for backup
  restore     Restore backup that was prepared using `shipperctl prepare` command

Flags:
  -f, --file string                         The path to a backup file (default "backup.yaml")
      --format string                       Output format. One of: json|yaml (default "yaml")
  -h, --help                                help for backup
      --kubeconfig string                   The path to the Kubernetes configuration file (default "~/.kube/config")
      --management-cluster-context string   The name of the context to use to communicate with the management cluster. defaults to the current one
  -v, --verbose                             Prints the list of backup items

Use "shipperctl backup [command] --help" for more information about a command.

$ shipperctl backup restore -v
Making sure shipper is down... done
Making sure there are no Shipper objects lying around... done
Would you like to see an overview of your backup? [y/n]: y
NAMESPACE  NAME                      OWNING APPLICATION
default    super-server-dc5bfc5a-0   super-server
default2   super-server2-dc5bfc5a-0  super-server2
default3   super-server3-dc5bfc5a-0  super-server3
Would you like to review backup? [y/n]: y
- application:
  ...
  releases:
  -  ...
...
Would you like to restore backup? [y/n]: y
...

$ shipperctl backup prepare -v
NAMESPACE  NAME                      OWNING APPLICATION
default    super-server-dc5bfc5a-0   super-server
default2   super-server2-dc5bfc5a-0  super-server2
default3   super-server3-dc5bfc5a-0  super-server3
Backup objects stored in "backup.yaml"

$ shipperctl backup prepare
Backup objects stored in "backup.yaml"

$ shipperctl backup prepare -f "a_new_bkup_file.json" --format json
Backup objects stored in "a_new_bkup_file.json"

## If shipper deoployment is not scaled down:
$ shipperctl backup restore
Making sure shipper is down... Error! shipper deployment has 1 replicas. Scale it down first with `kubectl -n shipper-system patch deploy shipper --type=merge -p '{"spec":{"replicas":0}}'`

## If there are shipper objects in the cluster:
$ shipperctl backup restore
Making sure shipper is down... done
Making sure there are no Shipper objects lying around... Error! found Shipper objects:
 - Applications: 3
 - Releases: 3
 - InstallationTargets: 3
 - TrafficTargets: 3
 - CapacityTargets: 3
delete them first with `kubectl delete app,rel,it,tt,ct --all-namespaces --all`

```

### DIFF
--- a/cmd/shipperctl/cmd/backup/backup.go
+++ b/cmd/shipperctl/cmd/backup/backup.go
@@ -32,9 +32,16 @@ var (
 	}
 )
 
-type shipperBackupObject struct {
-	Application shipper.Application `json:"application"`
-	Releases    []shipper.Release   `json:"releases"`
+type shipperBackupApplication struct {
+	Application    shipper.Application    `json:"application"`
+	BackupReleases []shipperBackupRelease `json:"backup_releases"`
+}
+
+type shipperBackupRelease struct {
+	Release            shipper.Release            `json:"release"`
+	InstallationTarget shipper.InstallationTarget `json:"installation_target"`
+	TrafficTarget      shipper.TrafficTarget      `json:"traffic_target"`
+	CapacityTarget     shipper.CapacityTarget     `json:"capacity_target"`
 }
 
 func init() {
@@ -61,22 +68,22 @@ func init() {
 	}
 }
 
-func marshalReleasesPerApplications(releasesPerApplications []shipperBackupObject) ([]byte, error) {
+func marshalReleasesPerApplications(releasesPerApplications []shipperBackupApplication) ([]byte, error) {
 	if outputFormat == "json" {
 		return json.MarshalIndent(releasesPerApplications, "", "    ")
 	}
 	return yaml.Marshal(releasesPerApplications)
 }
 
-func printReleasesPerApplications(releasesPerApplications []shipperBackupObject) {
+func printReleasesPerApplications(releasesPerApplications []shipperBackupApplication) {
 	tbl := table.New(
 		"NAMESPACE",
 		"RELEASE NAME",
 		"OWNING APPLICATION",
 	)
 	for _, obj := range releasesPerApplications {
-		for _, rel := range obj.Releases {
-
+		for _, backupRelease := range obj.BackupReleases {
+			rel := backupRelease.Release
 			tbl.AddRow(
 				rel.Namespace,
 				rel.Name,
--- a/cmd/shipperctl/cmd/backup/backup.go
+++ b/cmd/shipperctl/cmd/backup/backup.go
@@ -1,0 +1,89 @@
+package backup
+
+import (
+	"encoding/json"
+	"os"
+
+	"github.com/rodaine/table"
+	"github.com/spf13/cobra"
+	"sigs.k8s.io/yaml"
+
+	shipper "github.com/bookingcom/shipper/pkg/apis/shipper/v1alpha1"
+)
+
+var (
+	outputFormat             = "yaml"
+	kubeConfigFile           string
+	managementClusterContext string
+	verboseFlag              bool
+
+	BackupCmd = &cobra.Command{
+		Use:   "backup",
+		Short: "Backup and restore Shipper applications and releases",
+		PersistentPreRun: func(cmd *cobra.Command, args []string) {
+			switch outputFormat {
+			case "json", "yaml":
+				return
+			default:
+				cmd.Printf("error: output format %q not supported, allowed formats are: json, yaml\n", outputFormat)
+				os.Exit(1)
+			}
+		},
+	}
+)
+
+type shipperBackupObject struct {
+	Application shipper.Application `json:"application"`
+	Releases    []shipper.Release   `json:"releases"`
+}
+
+func init() {
+	kubeConfigFlagName := "kubeconfig"
+	fileFlagName := "file"
+
+	BackupCmd.PersistentFlags().StringVar(&kubeConfigFile, kubeConfigFlagName, "~/.kube/config", "The path to the Kubernetes configuration file")
+	if err := BackupCmd.MarkPersistentFlagFilename(kubeConfigFlagName, "yaml"); err != nil {
+		BackupCmd.Printf("warning: could not mark %q for filename autocompletion: %s\n", kubeConfigFlagName, err)
+	}
+
+	BackupCmd.PersistentFlags().StringVar(&managementClusterContext, "management-cluster-context", "", "The name of the context to use to communicate with the management cluster. defaults to the current one")
+	BackupCmd.PersistentFlags().BoolVarP(&verboseFlag, "verbose", "v", false, "Prints the list of backup items")
+	BackupCmd.PersistentFlags().StringVar(&outputFormat, "format", "yaml", "Output format. One of: json|yaml")
+
+	BackupCmd.PersistentFlags().StringVarP(&backupFile, fileFlagName, "f", "backup.yaml", "The path to a backup file")
+	err := BackupCmd.MarkPersistentFlagFilename(fileFlagName, "yaml")
+	if err != nil {
+		BackupCmd.Printf("warning: could not mark %q for filename yaml autocompletion: %s\n", fileFlagName, err)
+	}
+	err = BackupCmd.MarkPersistentFlagFilename(fileFlagName, "json")
+	if err != nil {
+		BackupCmd.Printf("warning: could not mark %q for filename json autocompletion: %s\n", fileFlagName, err)
+	}
+}
+
+func marshalReleasesPerApplications(releasesPerApplications []shipperBackupObject) ([]byte, error) {
+	if outputFormat == "json" {
+		return json.MarshalIndent(releasesPerApplications, "", "    ")
+	}
+	return yaml.Marshal(releasesPerApplications)
+}
+
+func printReleasesPerApplications(releasesPerApplications []shipperBackupObject) {
+	tbl := table.New(
+		"NAMESPACE",
+		"RELEASE NAME",
+		"OWNING APPLICATION",
+	)
+	for _, obj := range releasesPerApplications {
+		for _, rel := range obj.Releases {
+
+			tbl.AddRow(
+				rel.Namespace,
+				rel.Name,
+				obj.Application.Name,
+			)
+		}
+	}
+
+	tbl.Print()
+}
--- a/cmd/shipperctl/cmd/backup/backup.go
+++ b/cmd/shipperctl/cmd/backup/backup.go
@@ -68,20 +68,20 @@ func init() {
 	}
 }
 
-func marshalReleasesPerApplications(releasesPerApplications []shipperBackupApplication) ([]byte, error) {
+func marshalShipperBackupApplication(shipperBackupApplication []shipperBackupApplication) ([]byte, error) {
 	if outputFormat == "json" {
-		return json.MarshalIndent(releasesPerApplications, "", "    ")
+		return json.MarshalIndent(shipperBackupApplication, "", "    ")
 	}
-	return yaml.Marshal(releasesPerApplications)
+	return yaml.Marshal(shipperBackupApplication)
 }
 
-func printReleasesPerApplications(releasesPerApplications []shipperBackupApplication) {
+func printShipperBackupApplication(shipperBackupApplication []shipperBackupApplication) {
 	tbl := table.New(
 		"NAMESPACE",
 		"RELEASE NAME",
 		"OWNING APPLICATION",
 	)
-	for _, obj := range releasesPerApplications {
+	for _, obj := range shipperBackupApplication {
 		for _, backupRelease := range obj.BackupReleases {
 			rel := backupRelease.Release
 			tbl.AddRow(
--- a/cmd/shipperctl/cmd/backup/prepare.go
+++ b/cmd/shipperctl/cmd/backup/prepare.go
@@ -1,0 +1,101 @@
+package backup
+
+import (
+	"fmt"
+	"io/ioutil"
+	"strings"
+
+	"github.com/spf13/cobra"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/bookingcom/shipper/cmd/shipperctl/configurator"
+	"github.com/bookingcom/shipper/cmd/shipperctl/release"
+	"github.com/bookingcom/shipper/cmd/shipperctl/ui"
+)
+
+var (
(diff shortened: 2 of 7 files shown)


cmd/shipperctl/cmd/backup/backup.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/spf13/cobra"
+	"sigs.k8s.io/yaml"
+
+	shipper "github.com/bookingcom/shipper/pkg/apis/shipper/v1alpha1"
+)
+
+const (
+	kubeConfigFlagName = "kubeconfig"
+	fileFlagName       = "file"
+)
```
ArchangelX360, line 17: It feels like those declaration could be closer to their usage, I cannot find value into declaring them as package constants. Am I missing something?

cmd/shipperctl/cmd/backup/prepare.go
```diff
@@ -0,0 +1,86 @@
...
+		return err
+	}
+
+	namespaceList, err := kubeClient.CoreV1().Namespaces().List(metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+	var errList []string
+
+	releasesPerApplications := []releasesPerApplication{}
```
ArchangelX360, line 46: nit: semantically, this feels like it should be a map, it won't change much the code itself, but it might be slightly reduce the cognitively complexity. That is probably just a personal preference.

cmd/shipperctl/cmd/backup/prepare.go
```diff
@@ -0,0 +1,86 @@
...
+		errList = append(errList, err.Error())
+	} else {
+		err := ioutil.WriteFile(backupFile, data, 0644)
+		if err != nil {
+			errList = append(errList, err.Error())
+		}
+	}
+
+	if len(errList) > 0 {
+		return fmt.Errorf(strings.Join(errList, ", "))
```
ArchangelX360, line 81: If there is no error at `printAndUnmarshalReleasesPerApplications`, but errors when retrieving some releases, the file will be created, but the command will crash with errors. Is that expected?

What about either:
- emit warnings when failing to retrieve releases of an application, and crashing if the `printAndUnmarshalReleasesPerApplications` is failing
or
- crashing at L57 to the first faill to retrieve releases of an application

ArchangelX360 (changes requested):

cmd/shipperctl/cmd/backup/prepare.go
hihilla: Solid point
How about if there were errors retrieving objects, warn user and ask to continue or stop? And if user chooses to continue, emit those warnings

ArchangelX360: That works for me!

cmd/shipperctl/cmd/backup/restore.go
```diff
@@ -0,0 +1,201 @@
...
+		return err
+	}
+	if confirm {
+		printReleasesPerApplications(releasesPerApplications)
+		confirm, err = ui.AskForConfirmation(cmd.InOrStdin(), "Would you like to review backup?")
+		if err != nil {
+			return err
+		}
+		if confirm {
+			data, err := marshalReleasesPerApplications(releasesPerApplications)
```
nodo, line 67: What's the reason we asked again to print the releases here?

cmd/shipperctl/cmd/backup/restore.go
```diff
@@ -0,0 +1,201 @@
...
+				errList = append(errList, err.Error())
+				continue
+			}
+			fmt.Printf("release %q owner reference updates with uid %q\n", fmt.Sprintf("%s/%s", rel.Namespace, rel.Name), *uid)
+
+			rel.ResourceVersion = ""
+			if _, err = shipperClient.ShipperV1alpha1().Releases(rel.Namespace).Create(rel); err != nil {
+				errList = append(errList, err.Error())
+			}
+			fmt.Printf("release %q created\n", fmt.Sprintf("%s/%s", rel.Namespace, rel.Name))
```
nodo, line 122: Do you foresee any problem if _some_ of the releases fail to be restored? (not a blocker for now, just a thought)

cmd/shipperctl/cmd/backup/restore.go
```diff
@@ -0,0 +1,201 @@
...
+		}
+
+	}
+	if len(errList) > 0 {
+		return fmt.Errorf(strings.Join(errList, ", "))
+	}
+	return nil
+}
```
nodo, line 130: Really like this function, super clear!

cmd/shipperctl/cmd/backup/restore.go
hihilla: The first is an overview: a table of the releases and their application
The second is the backup file itself

cmd/shipperctl/cmd/backup/restore.go
hihilla: Some of the releases fail to be restored == loss of history.
I don't know if this is the best way to overcome this. I think in this scenario I would want to apply the backup again, but it would fail to create the application (already there error) and would skip its releases.
I;m not sure what is the best solution here.

nodo: What about if the "last 3" fails to restore. Would that cause problems? or just restore that particular version?

cmd/shipperctl/cmd/backup/restore.go
nodo: 👍

cmd/shipperctl/cmd/backup/restore.go
hihilla: Yeah that's way worse: if the latest release(s) fails to restore, the live version is not the latest. 
We should probably panic upon first error instead of collecting them. WDYT?

nodo: Yes, that would neat!

cmd/shipperctl/cmd/backup/restore.go
```diff
@@ -89,47 +88,40 @@ func runRestoreCommand(cmd *cobra.Command, args []string) error {
...
 
-func restore(releasesPerApplications []shipperBackupObject, kubeClient kubernetes.Interface, shipperClient versioned.Interface) error {
-	var errList []string
+func restore(releasesPerApplications []shipperBackupObject, kubeClient kubernetes.Interface, shipperClient shipperclientset.Interface) error {
 	for _, obj := range releasesPerApplications {
 		// apply application
 		if err := applyApplication(kubeClient, shipperClient, obj.Application); err != nil {
-			errList = append(errList, err.Error())
-			continue
+			return err
```
nodo, line 95: Wouldn't this stop the `restore` on every error? (as oppose to stop only on the first error)?

hihilla: If `applyApplication` fails, `restore` will throw the same error, which will be the returned error by the command.
The first error encountered by the restoring process will stop the command entirely.
I think this is better since, if the restore only partially worked, I would delete all restored objects and try again

nodo: Yes, totally understand that.

But I thought that we were going for a different approach:
- Either fail if the incumbent release fails, or
- Keep going and collect of the errors if the "history" release fails

I think in this case we could be in a situation where an old release comes back to life 🧟

hihilla: I'm a bit confused
> I think in this case we could be in a situation where an old release comes back to life 🧟

this case == panic out on every error?

> But I thought that we were going for a different approach:

I would assume that in case I am trying to restore a backup and I have failed to restore some of the history releases (the only error I have seen so far is "already exists", so I don't know which errors can occur), I wouldn't be happy with this situation and I would delete everything and restart the restore from scratch (knowing that shipper is done so I'm not loosing any new data).

I guess I should clarify on the usage of `restore` that shipper must be down during restore...

nodo: Just to wrap this up for future reference, we agree that any time there is an error we bail because we want to restore _all_ the data, history is as important as the incumbent release. If there is an error the operator should clean up the partially restored objects and retry.

cmd/shipperctl/cmd/backup/backup.go
```diff
@@ -0,0 +1,96 @@
...
+	if err != nil {
+		BackupCmd.Printf("warning: could not mark %q for filename yaml autocompletion: %s\n", fileFlagName, err)
+	}
+	err = BackupCmd.MarkPersistentFlagFilename(fileFlagName, "json")
+	if err != nil {
+		BackupCmd.Printf("warning: could not mark %q for filename json autocompletion: %s\n", fileFlagName, err)
+	}
+}
+
+func marshalReleasesPerApplications(releasesPerApplications []shipperBackupApplication) ([]byte, error) {
```
ArchangelX360: Now that the code is using `shipperBackupApplication` type, could we rename `releasesPerApplications` parameter name and maybe also strip it out of the function name `marshalReleasesPerApplications`?
(this is applicable to `printReleasesPerApplications`  function as well, and maybe other places in the code)

cmd/shipperctl/cmd/backup/restore.go
```diff
@@ -157,24 +220,33 @@ func uidOfApplication(shipperClient shipperclientset.Interface, appName, appName
...
-	if n := len(rel.OwnerReferences); n != 1 {
-		key, _ := cache.MetaNamespaceKeyFunc(rel)
-		return nil, fmt.Errorf("expected exactly one owner for Release %q but got %d", key, n)
+func uidOfRelease(shipperClient shipperclientset.Interface, relName, relNamespace string) (*types.UID, error) {
+	release, err := shipperClient.ShipperV1alpha1().Releases(relNamespace).Get(relName, metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+	uid := release.GetUID()
+	return &uid, nil
```
nodo, line 229: What is the reason you decided to return a `*types.UID` instead of `types.UID`?

nodo (commented):
How are would it be to add unit tests?

cmd/shipperctl/cmd/backup/restore.go
ArchangelX360: The `restore` command needs two pre-steps:
- Shipper must be down
- Shipper objects must be deleted from the cluster

Crashing with an "Application already exists" message to the operator restoring the backup could be confusing on what would be the next thing the operator must do.
Adding to that the stress level of an operator restoring a Shipper backup after a procedural error was made, and this could be slowing down a lot the process of recovery.

What I suggest is to actually implement verifiers of those pre-steps at the beginning of the `restore` command.
- "check if Shipper is up, crash if not with a very informative message on what action should be done by the operator"
- "check if there is some Shipper object in the cluster, crash if not with a very informative message on what action should be done by the operator"

WDYGT?

cmd/shipperctl/cmd/backup/restore.go
hihilla: That's a good question that I have no answer to. This probably shouldn't be a pointer

hihilla:
> How are would it be to add unit tests?

I am not 100% sure how to test this without a fake client, and even with, I'm not sure how to test this...

cmd/shipperctl/cmd/backup/restore.go
hihilla: @ArchangelX360  @nodo 
For the second point: if some shipper objects were found in the cluster, should the command offer to delete them? Or only bail out with "please clean" message?

ArchangelX360: Bailing out I think, keep the command dumb is usually a good thing in under stress operation workflow.

Ideally, the operator should follow a list of instructions that mentions:
1. take down Shipper by running x, y, z
2. delete Shipper objects by running x, y, z
3. restore backup by running `shipperctl backup restore`
These checks I am suggesting here is to prevent a "I skipped an instruction" situation.

In the future could decide to make step one and 2 part of `shipperctl backup restore` in the future to have a fully automated procedure with manual confirmation prompt. This does not mean that the command will be smart, it won't recover by itself, but taking the correct step in the same order, which is less confusing.

I think for the sake of this first MR, we can stay simple and not implement step 1 and 2, just checking that they are done by the operator.

nodo: I would rather not delete the object during restore, I think it might be a very unexpected behaviour.

hihilla: Even with a prompt to ask about it before?
And what about scaling down shipper? Should the command (ask to do it and then) do it? Or only verify?

nodo: Yeah, just verify and just bail out with an error in both cases. 

The rationale behind is
- deleting objects is easy to do with `kubectl` anyway
- adds complexity to these commands
- During a restore if you see a unwanted object or a problem you might want to take a step back and check more manual checks

nodo:
Adding a fake client is exactly the wait to go, maybe you could try and use https://pkg.go.dev/k8s.io/client-go/kubernetes/fake for a fake k8s client. I am not 100% sure on how to do it but I think it would be worth, let me know if I can help out somehow.

cmd/shipperctl/cmd/backup/restore.go
hihilla: expected output for the 2 checks:

```
## If shipper deoployment is not scaled down:
$ shipperctl backup restore
Making sure shipper is down... Error! shipper deployment has 1 replicas. Scale it down first with `kubectl -n shipper-system patch deploy shipper --type=merge -p '{"spec":{"replicas":1}}'`

## If there are shipper objects in the cluster:
$ shipperctl backup restore
Making sure shipper is down... done
Making sure there are no Shipper objects lying around... Error! found Shipper objects:
 - Applications: 3
 - Releases: 3
 - InstallationTargets: 3
 - TrafficTargets: 3
 - CapacityTargets: 3
delete them first with `kubectl delete app,rel,it,tt,ct --all-namespaces --all`
```

nodo (approved):
🎉

ArchangelX360 (approved):

nikoloup (approved):

PR #378, opened 2021-01-05
WIP: fix moving back in strategy bug

hihilla:
When a current release is progressing shipper will:
1. increase capacity to current (or *contender*)
2. increase traffic to current
3. decrease traffic to previous (or *incumbent*)
4. decrease capacity to previous

When a release is moving back in the strategy (not progressing), this process will cause pods dropping.
So when a release in **not** progressing, shipper will first move capacity and traffic to the previous release and then reduce traffic and capacity from current release.


pkg/controller/release/strategy_executor.go
```diff
@@ -147,21 +149,47 @@ func (e *StrategyExecutor) Execute(prev, curr, succ *releaseInfo) (bool, []Strat
...
+				pipeline.Enqueue(genCapacityEnforcer(prevctx, prev, curr))
+			}
+		} else {
+			pipeline.Enqueue(genTrafficEnforcer(ctx, curr, succ))
+			pipeline.Enqueue(genCapacityEnforcer(ctx, curr, succ))
 		}
 	} else {
-		pipeline.Enqueue(genTrafficEnforcer(ctx, curr, succ))
-		pipeline.Enqueue(genCapacityEnforcer(ctx, curr, succ))
+		// release is not progressing (target step > achieved step):
```
ArchangelX360, line 175: You probably meant: "// release is not progressing (target step < achieved step)"

pkg/controller/release/strategy_executor.go
```diff
@@ -147,21 +149,47 @@ func (e *StrategyExecutor) Execute(prev, curr, succ *releaseInfo) (bool, []Strat
...
-		if hasTail {
-			// This is the moment where a contender is performing a look-behind.
-			// Incumbent's context is completely identical to it's successor
-			// except that it's not the head of the chain anymore.
-			prevctx := ctx.Copy()
-			prevctx.isHead = false
-			pipeline.Enqueue(genTrafficEnforcer(prevctx, prev, curr))
-			pipeline.Enqueue(genCapacityEnforcer(prevctx, prev, curr))
+	if e.progressing {
+		// progressing release (target step > achieved step):
```
ArchangelX360, line 153: You probably meant: "// release is not progressing (target step >= achieved step)"

pkg/util/release/environment_test.go
```diff
@@ -0,0 +1,54 @@
...
+		},
+		{
+			"not progressing release",
+			&shipper.AchievedStep{
+				Step: 1,
+				Name: "1",
+			},
+			0,
+			false,
+		},
```
ArchangelX360, line 45: Could you add a test for the equality? (achieved step == targetStep) As it has been a confusion in the comments of the code, I think clearly stating the intent of the code in tests could alleviate some doubt from the reader.

ArchangelX360 (changes requested):

pkg/util/release/environment_test.go
```diff
@@ -0,0 +1,63 @@
...
+			"progressing release",
+			&shipper.AchievedStep{
+				Step: 0,
+				Name: "0",
+			},
+			2,
+			true,
+		},
+		{
+			"progressing release (target step == achieved step)",
```
ArchangelX360, line 38: Why is the release considered to be progressing when the two steps are equal?

PR #380, opened 2021-01-08
Add resources limits and requests to the deployment manifests

hihilla:
This states the resources Shipper requiers, incase there is a resource manager of some sort set in place.


Makefile
```diff
@@ -177,6 +185,19 @@ IMAGE_NAME_TO_USE = $(if $(USE_IMAGE_NAME_WITH_SHA256),$(IMAGE_NAME_WITH_SHA256)
 build/%.deployment.$(IMAGE_TAG).yaml: kubernetes/%.deployment.yaml build/%.image.$(IMAGE_TAG) build
 	sed s=\<IMAGE\>=$(IMAGE_NAME_TO_USE)= $< > $@
 
+# Build deployment.yaml files for e2e tests.
+# These manifests don't define resource requests and limits.
+IMAGE_NAME_WITH_SHA256 = $(shell cat build/$*.image.$(IMAGE_TAG))
+IMAGE_NAME_TO_USE = $(if $(USE_IMAGE_NAME_WITH_SHA256),$(IMAGE_NAME_WITH_SHA256),$(IMAGE_NAME_WITH_TAG))
```
nodo, line 191: Lines 190 and 191 are copied from lines 183 and 184 is it expected?

hihilla: yes, but Titouan as suggested to utilise kustomize for this, which I think is a better solution

nikoloup (approved):

PR #387, opened 2021-01-27
Fix dropping pods when moving back in the strategy

hihilla:
Introducing the notion of steppingBackwards to the strategy executor

When a current release is not steppingBackwards (standing or moving forward in the strategy - the most common states) shipper will:

1. increase capacity to current
2. increase traffic to current
3. decrease traffic to previous
4. decrease capacity to previous

When a release is moving back in the strategy (steppingBackwards), this process will cause pods dropping.
So when a release in *steppingBackwards*, shipper will first move capacity and traffic to the previous release and then reduce traffic and capacity from current release.

### DIFF
--- a/pkg/controller/release/release_controller.go
+++ b/pkg/controller/release/release_controller.go
@@ -409,12 +409,13 @@ func (c *Controller) executeReleaseStrategy(relinfo *releaseInfo, diff *diffutil
 	}
 
 	// choose strategy and step and execute
-	strategy, targetStep, err := strategyAndStepToExecute(rel, relinfoSucc)
+	strategy, targetStep, isSteppingBackwards, err := strategyAndStepToExecute(rel, relinfoSucc)
 	if err != nil {
 		return nil, nil, err
 	}
-	executor := NewStrategyExecutor(strategy, targetStep)
-	complete, patches, trans := executor.Execute(relinfoPrev, relinfo, relinfoSucc)
+	executor := NewStrategyExecutor(strategy, targetStep, isSteppingBackwards)
+	complete, patches, trans := executor.
+		Execute(relinfoPrev, relinfo, relinfoSucc, NewExecutorPipeline())
 
 	// update logs, conditions, events
 	if len(patches) == 0 {
@@ -500,7 +501,7 @@ func (c *Controller) recordEvents(rel *shipper.Release, achievedStep int32, prev
 	}
 }
 
-func strategyAndStepToExecute(rel *shipper.Release, relinfoSucc *releaseInfo) (*shipper.RolloutStrategy, int32, error) {
+func strategyAndStepToExecute(rel *shipper.Release, relinfoSucc *releaseInfo) (*shipper.RolloutStrategy, int32, bool, error) {
 	var succ *shipper.Release
 	if relinfoSucc != nil {
 		succ = relinfoSucc.release
@@ -508,23 +509,26 @@ func strategyAndStepToExecute(rel *shipper.Release, relinfoSucc *releaseInfo) (*
 	isHead := succ == nil
 	var strategy *shipper.RolloutStrategy
 	var targetStep int32
+	var isSteppingBackwards bool
 	// A head release uses it's local spec-defined strategy, any other release
 	// follows it's successor state, therefore looking into the forecoming spec.
 	if isHead {
 		strategy = rel.Spec.Environment.Strategy
 		targetStep = rel.Spec.TargetStep
+		isSteppingBackwards = releaseutil.IsReleaseSteppingBackwards(rel.Status.AchievedStep, rel.Spec.TargetStep)
 	} else {
 		strategy = succ.Spec.Environment.Strategy
 		targetStep = succ.Spec.TargetStep
+		isSteppingBackwards = releaseutil.IsReleaseSteppingBackwards(succ.Status.AchievedStep, succ.Spec.TargetStep)
 	}
 
 	// Looks like a malformed input. Informing about a problem and bailing out.
 	if targetStep >= int32(len(strategy.Steps)) {
 		err := fmt.Errorf("no step %d in strategy for Release %q",
 			targetStep, controller.MetaKey(rel))
-		return nil, 0, shippererrors.NewUnrecoverableError(err)
+		return nil, 0, true, shippererrors.NewUnrecoverableError(err)
 	}
-	return strategy, targetStep, nil
+	return strategy, targetStep, isSteppingBackwards, nil
 }
 
 func (c *Controller) buildReleasesInfo(relinfo *releaseInfo) (*releaseInfo, *releaseInfo, error) {
--- a/pkg/controller/release/strategy_executor.go
+++ b/pkg/controller/release/strategy_executor.go
@@ -39,17 +39,32 @@ func (ctx *context) Copy() *context {
 
 type PipelineStep func(shipper.RolloutStrategyStep, conditions.StrategyConditionsMap) (PipelineContinuation, []StrategyPatch, []ReleaseStrategyStateTransition)
 
-type Pipeline []PipelineStep
+type Pipeline interface {
+	Enqueue(step PipelineStep)
+	Process(strategyStep shipper.RolloutStrategyStep, cond conditions.StrategyConditionsMap) (bool, []StrategyPatch, []ReleaseStrategyStateTransition)
 
-func NewPipeline() *Pipeline {
-	return new(Pipeline)
+	/*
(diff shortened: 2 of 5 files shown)


pkg/controller/release/release_controller.go
```diff
@@ -460,7 +460,8 @@ func (c *Controller) executeReleaseStrategy(relinfo *releaseInfo, diff *diffutil
 		return nil, nil, shippererrors.NewUnrecoverableError(err)
 	}
 
-	executor := NewStrategyExecutor(strategy, targetStep)
+	steppingBackwards := releaseutil.IsReleaseSteppingBackwards(rel.Status.AchievedStep, rel.Spec.TargetStep)
```
nikoloup, line 463: Perhaps we could rename this to `isSteppingBackwards`?

pkg/controller/release/strategy_executor.go
```diff
@@ -147,7 +149,30 @@ func (e *StrategyExecutor) Execute(prev, curr, succ *releaseInfo) (bool, []Strat
...
+				pipeline.Enqueue(genCapacityEnforcer(prevctx, prev, curr))
+				pipeline.Enqueue(genTrafficEnforcer(prevctx, prev, curr))
+			}
+			pipeline.Enqueue(genTrafficEnforcer(ctx, curr, succ))
+			pipeline.Enqueue(genCapacityEnforcer(ctx, curr, succ))
+		} else {
+			pipeline.Enqueue(genTrafficEnforcer(ctx, curr, succ))
+			pipeline.Enqueue(genCapacityEnforcer(ctx, curr, succ))
+		}
+	} else if isHead {
```
nikoloup, line 175: Maybe it's just me, but it would be nice to have the same comment here as above with the steps in the forward order?

pkg/controller/release/strategy_executor.go
```diff
@@ -147,7 +149,30 @@ func (e *StrategyExecutor) Execute(prev, curr, succ *releaseInfo) (bool, []Strat
...
+		// increase capacity for previous release
+		// increase traffic for previous release
+		// reduce traffic for current release
+		// reduce capacity for current release
+		if isHead {
+			if hasTail {
+				prevctx := ctx.Copy()
+				prevctx.isHead = false
+				pipeline.Enqueue(genCapacityEnforcer(prevctx, prev, curr))
+				pipeline.Enqueue(genTrafficEnforcer(prevctx, prev, curr))
```
nikoloup, line 167: This is definitely out of the scope of this MR, but for the future: I find this snippet quite confusing. As far as I understand, the order of Capacity and Traffic enforcing is something that depends entirely on whether you are "increasing" or "decreasing". If that is the case and there's no other case I'm ignoring here, I feel an abstraction of the kind "increaseForRelease" or something else aptly named that takes care of the correct order of enforcements would contribute to clarity here. Otherwise this can be tricky to read and can potentially lead to mistakes when working on this part.

nikoloup (approved):
LGTM!

pkg/controller/release/strategy_executor.go
hihilla: You are seeing one of the reasons I would like to make changes to this code. I agree with you, if we look at this as a state machine, we should be able to construct the notion of "increasing" or "decreasing" for release. However, the strategy executor is not implemented as a state machine, but as a pipeline.

nikoloup: Ah, right, that would definitely be ideal - but even without that, could we not have another method that enqueues the actions to the pipeline in the correct order depending on whether we are increasing or decreasing?

hihilla: I don't think I fully understand what you mean.
I think somewhere something in shipper needs to know the direction of the progress of a given release, in order to do this in the correct order. 
Generally speaking, Shipper should be able to look at a release and know what exactly to enforce on it. All the data about this release (if it's the contender, if it's increasing or decreasing, and if it needs capacity or traffic enforcing) is already in the body of the release (spec and status)

pkg/controller/release/strategy_executor.go
```diff
@@ -147,7 +149,30 @@ func (e *StrategyExecutor) Execute(prev, curr, succ *releaseInfo) (bool, []Strat
...
+			}
+			pipeline.Enqueue(genTrafficEnforcer(ctx, curr, succ))
+			pipeline.Enqueue(genCapacityEnforcer(ctx, curr, succ))
+		} else {
+			pipeline.Enqueue(genTrafficEnforcer(ctx, curr, succ))
+			pipeline.Enqueue(genCapacityEnforcer(ctx, curr, succ))
```
nodo, line 173: Minor:

```go
pipeline.Enqueue(genTrafficEnforcer(ctx, curr, succ))
pipeline.Enqueue(genCapacityEnforcer(ctx, curr, succ))
```
Happens in both `if` and `else`, we should just have it at the end of the block outside the `if` something like:


```go
		if isHead && hasTail {
			prevctx := ctx.Copy()
			prevctx.isHead = false
			pipeline.Enqueue(genCapacityEnforcer(prevctx, prev, curr))
			pipeline.Enqueue(genTrafficEnforcer(prevctx, prev, curr))
		}
		pipeline.Enqueue(genTrafficEnforcer(ctx, curr, succ))
		pipeline.Enqueue(genCapacityEnforcer(ctx, curr, succ))
```

pkg/controller/release/strategy_executor.go
```diff
@@ -147,7 +149,30 @@ func (e *StrategyExecutor) Execute(prev, curr, succ *releaseInfo) (bool, []Strat
...
+		}
```
nodo, line 174: It seems that the block:
```go
			pipeline.Enqueue(genTrafficEnforcer(ctx, curr, succ))
			pipeline.Enqueue(genCapacityEnforcer(ctx, curr, succ))
```
is the same in both branches of the `if` statement, could we have it just once?

E.g.

```go
		if isHead && hasTail {
			prevctx := ctx.Copy()
			prevctx.isHead = false
			pipeline.Enqueue(genCapacityEnforcer(prevctx, prev, curr))
			pipeline.Enqueue(genTrafficEnforcer(prevctx, prev, curr))
		}
		pipeline.Enqueue(genTrafficEnforcer(ctx, curr, succ))
		pipeline.Enqueue(genCapacityEnforcer(ctx, curr, succ))
```

pkg/controller/release/strategy_executor.go
nikoloup: Hmm, I'm probably misunderstanding something. What I meant was literally:
```
func increase(ctx, rel, rel) {
  pipeline.Enqueue(genCapacityEnforcer(ctx, rel, rel))
  pipeline.Enqueue(genTrafficEnforcer(ctx, rel, rel))
}
func decrease(ctx, rel, rel) {
  pipeline.Enqueue(genTrafficEnforcer(ctx, rel, rel))
  pipeline.Enqueue(genCapacityEnforcer(ctx, rel, rel))
}
```

hihilla: Ah ok now I understand what you mean.
Yeah that might make things easier to read.

nikoloup: Sorry, I didn't explain it really clearly at the beginning 😅

nikoloup (dismissed):

pkg/controller/release/strategy_executor.go
```diff
@@ -144,24 +171,47 @@ func (e *StrategyExecutor) Execute(prev, curr, succ *releaseInfo) (bool, []Strat
 		isHead:     isHead,
 	}
 
-	pipeline := NewPipeline()
+	//pipeline := NewExecutorPipeline()
```
ArchangelX360, line 174: nit: you should delete that now :D

pkg/controller/release/strategy_executor_test.go
```diff
@@ -0,0 +1,339 @@
...
+const succRelUid = "random-succ-uid"
+const currRelUid = "random-curr-uid"
+const prevRelUid = "random-prev-uid"
+
+func TestExecute(t *testing.T) {
+	var tests = []struct {
+		name                string
+		prev, curr, succ    *releaseInfo
+		isSteppingBackwards bool
+		expectedOut         string
```
ArchangelX360, line 63: Could we make this an `[]string` instead? It will be more readable in test cases IMO.

ArchangelX360 (changes requested):

nikoloup (dismissed):
LGTM!

ArchangelX360 (approved):

PR #392, opened 2021-04-07
Webhook validates deletion

hihilla:
This adds Delete action to validating webhook. 
This is necessary for when there's a rollout block and a user is not aware. In this case, the user might delete the contender, but shipper will not reactivate the incumbent.

Shipper will also populate override annotations from releases to owning applications, to prevent a case where an overriding release is deleted, but the application is not overriding the block so Shipper will not reactivated incumbent release.

### DIFF
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -220,18 +220,39 @@ func (c *Webhook) validateHandlerFunc(review *admission.AdmissionReview) *admiss
 	request := review.Request
 	var err error
 
+	if request.Operation == kubeclient.Delete {
+		err = c.validateDelete(request)
+	} else {
+		err = c.validateCreateUpdate(request)
+	}
+
+	if err != nil {
+		return &admission.AdmissionResponse{
+			Result: &metav1.Status{
+				Message: err.Error(),
+			},
+		}
+	}
+
+	return &admission.AdmissionResponse{
+		Allowed: true,
+	}
+}
+
+func (c *Webhook) validateCreateUpdate(request *kubeclient.AdmissionRequest) error {
+	var err error
 	switch request.Kind.Kind {
 	case "Application":
 		var application shipper.Application
 		err = json.Unmarshal(request.Object.Raw, &application)
 		if err == nil {
-			err = c.validateApplication(request, application)
+			err = c.validateBlocksForApplication(request, application)
 		}
 	case "Release":
 		var release shipper.Release
 		err = json.Unmarshal(request.Object.Raw, &release)
 		if err == nil {
-			err = c.validateRelease(request, release)
+			err = c.validateBlocksForRelease(request, release)
 		}
 	case "Cluster":
 		var cluster shipper.Cluster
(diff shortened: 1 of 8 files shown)


pkg/webhook/webhook.go
```diff
@@ -249,18 +270,28 @@ func (c *Webhook) validateHandlerFunc(review *admission.AdmissionReview) *admiss
...
+	default:
+		return nil
 	}
-
-	return &admission.AdmissionResponse{
-		Allowed: true,
+	if err == nil {
+		err = c.validateObject(obj)
 	}
+	return err
```
ArchangelX360, line 294:
```suggestion
	return c.validateObject(obj)
```

pkg/webhook/webhook.go
```diff
@@ -322,3 +353,19 @@ func (c *Webhook) validateApplication(request *admission.AdmissionRequest, appli
...
+		return err
+	}
+	if err = rolloutblock.ValidateAnnotations(existingBlocks, overrides); err != nil {
+		return err
+	}
+	if err = rolloutblock.ValidateBlocks(existingBlocks, overrides); err != nil {
+		return err
+	}
+
+	return err
```
ArchangelX360, line 370:
```suggestion
	return nil
```

pkg/webhook/webhook.go
```diff
@@ -322,3 +353,19 @@ func (c *Webhook) validateApplication(request *admission.AdmissionRequest, appli
 
 	return err
 }
+
+func (c *Webhook) validateObject(obj metav1.Object) error {
+	var err error
```
ArchangelX360, line 358:
```suggestion
```

pkg/webhook/webhook.go
```diff
@@ -322,3 +353,19 @@ func (c *Webhook) validateApplication(request *admission.AdmissionRequest, appli
...
+	overrides, existingBlocks, err := rolloutblock.GetAllBlocks(c.rolloutBlocksLister, obj)
+	if err != nil {
+		return err
+	}
+	if err = rolloutblock.ValidateAnnotations(existingBlocks, overrides); err != nil {
```
ArchangelX360, line 363:
```suggestion
	if err := rolloutblock.ValidateAnnotations(existingBlocks, overrides); err != nil {
```

pkg/webhook/webhook.go
```diff
@@ -322,3 +353,19 @@ func (c *Webhook) validateApplication(request *admission.AdmissionRequest, appli
...
+		return err
+	}
+	if err = rolloutblock.ValidateBlocks(existingBlocks, overrides); err != nil {
```
ArchangelX360, line 366:
```suggestion
	if err := rolloutblock.ValidateBlocks(existingBlocks, overrides); err != nil {
```

pkg/webhook/webhook.go
```diff
@@ -322,3 +353,19 @@ func (c *Webhook) validateApplication(request *admission.AdmissionRequest, appli
 
 	return err
 }
+
+func (c *Webhook) validateObject(obj metav1.Object) error {
```
ArchangelX360, line 357: Could the function name mention something about blocks? Or this method being pushed to the `rolloutblock` package? As per the signature, it does not feel like it would only check for block related things.

ArchangelX360 (changes requested):

nikoloup (approved):
LGTM!

ArchangelX360 (approved):

PR #394, opened 2021-04-13
Add flag InClusterconfig to prepare backup command

hihilla:
This flag will be used when the prepare backup command is used within a pod in a reoccurring job

### DIFF
--- a/cmd/shipperctl/cmd/backup/prepare.go
+++ b/cmd/shipperctl/cmd/backup/prepare.go
@@ -8,14 +8,18 @@ import (
 	"github.com/spf13/cobra"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
 
 	"github.com/bookingcom/shipper/cmd/shipperctl/config"
+	"github.com/bookingcom/shipper/cmd/shipperctl/configurator"
 	"github.com/bookingcom/shipper/cmd/shipperctl/release"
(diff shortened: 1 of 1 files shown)


cmd/shipperctl/config/kubeconfig.go
```diff
@@ -25,3 +26,15 @@ func Load(kubeConfigFile, managementClusterContext string) (kubernetes.Interface
...
+func LoadInClusterConfig(restConfig *rest.Config) (kubernetes.Interface, shipperclientset.Interface, error) {
+	kubeClient, err := configurator.NewKubeClientFromInClusterKubeConfig(restConfig)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	shipperClient, err := configurator.NewShipperClientFromInClusterKubeConfig(restConfig)
+	if err != nil {
+		return nil, nil, err
+	}
```
ArchangelX360, line 38: Why not using `NewClusterConfigurator(restConfig)` directly here and delete `NewKubeClientFromInClusterKubeConfig` and `NewShipperClientFromInClusterKubeConfig`?

ArchangelX360 (changes requested):

cmd/shipperctl/config/kubeconfig.go
ArchangelX360: Actually, even removing this function and just use `NewClusterConfigurator` in the upper scope in place of `LoadInClusterConfig`.

ArchangelX360 (dismissed):

ArchangelX360 (approved):

PR #396, opened 2021-05-10
Installation controller fails to create when a namespace is terminating

hihilla:
This change makes the kubeclient error with cause `Namespace Terminating`, or a forbidden (404) error with kind `namespaces` (namespace not found error) retryable

### DIFF
--- a/docs/start/install.rst
+++ b/docs/start/install.rst
@@ -1,7 +1,7 @@
 .. _start:
 
 ####################
-Shipper in 5 minutes
+Installing Shipper
 ####################
 
 *************************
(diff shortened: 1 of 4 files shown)


ArchangelX360 (approved):

docs/start/install.rst
```diff
@@ -170,6 +170,21 @@ Step 6: do a rollout!
 
 Now you should have a working Shipper installation. :ref:`Let's roll something out! <user_rolling-out>`
 
+
+*****************
+Namespace manager
+*****************
+By design, Shipper does not create namespaces in the application cluster.
+In case the namespace does not exist in the application cluster, and this application cluster is selected for a *Release*,
```
nikoloup, line 178: Should we specify here just for completeness that the expected namespace in the application cluster has to be of the same name?

nikoloup (approved):

docs/start/install.rst
hihilla: hmm you're right

nikoloup (approved):

PR #402, opened 2021-12-03
Create explicit Shipper application cluster role

nikoloup:
This MR changes shipperctl so that it creates an explicit application-cluster role with fewer permissions. Previously shipperctl would give shipper cluster admin rights to the application clusters.


farhatmihalko:
LGTM!

farhatmihalko (approved):

